PR #9, opened 2020-06-01
2042 relax template switch

sanjuroj:


js/base.js
```diff
@@ -706,7 +706,7 @@ window.create_unified_changeset_notice = ({
     })
 }
 
-const createPnotifyAlert = (passedInConfig) => {
+const   createPnotifyAlert = (passedInConfig) => {
```
camgnostic, line 709: what changed here?

sanjuroj: I'm betting that was actually a fat finger.   Fixed.

PR #10, opened 2020-06-02
1595 user admin admin filter

camgnostic:
Closes mercycorps/TolaActivity#1595


tola_management/views.py
```diff
@@ -502,181 +503,57 @@ def get_list_queryset(self):
         req = self.request
 
         #theres a bug with django rest framework pagination that prevents this from working
```
sanjuroj, line 505: Is this comment still needed?

PR #13, opened 2020-06-03
Unified changeset pnotify wrapper

axoplasm:


tola/urls.py
```diff
@@ -58,6 +58,9 @@
 urlpatterns += [
     url(r'^jsi18n/$', JavaScriptCatalog.as_view(), name='javascript-catalog'),
 
+    # pnotify prototype
+    url(r'^pnotify-prototype/', TemplateView.as_view(template_name='prototypes/pnotify-prototype.html')),
```
sanjuroj, line 62: Are you leaving this in place on purpose for now, and planning to take it out before we merge into dev?

axoplasm: sure. Let me mark it with a TODO

PR #39, opened 2020-07-06
2394 new pnotify styling

camgnostic:
Applies new styling to multiple entry points for changeset function


js/components/changesetNotice.js
```diff
@@ -151,24 +151,16 @@ const create_unified_changeset_notice = ({
...
-                    is_valid = false;
+            switch (validation_type) {
+                case 1:
+                    // Uses RFC dropdown logic (either a rationale or a non-Other reason for change required):
+                    is_valid = (rationale || (reasons_for_change.length > 0 && reasons_for_change.indexOf(1) == -1))
+                    break;
+                case 0:
+                default:
+                    // Either a rationale is submitted, or there was no rationale form, or it was optional:
+                    is_valid = (rationale || !include_rationale || !rationale_required);
```
sanjuroj, line 162: I wonder if the default option needs to be a little more complex.  Should it check both the `rationale_required` and the `rfc_required` fields and their corresponding values in the form?

camgnostic: Definitely needs to if we aren't going to remove rfc_required as an option.  I'm thinking maybe after we get everything moved over to the unified function we can wean ourselves off of a couple of those parameters (like, maybe RFC dropdown is complex enough to require the "validation_type" parameter and that simplifies some of it?)

PR #75, opened 2020-07-28
2337 user program roles nesting

camgnostic:
Adds expando/collapso behavior to countries in the user admin per mercycorps/TolaActivity#2337

### DIFF
--- a/js/pages/tola_management_pages/user/components/edit_user_programs.js
+++ b/js/pages/tola_management_pages/user/components/edit_user_programs.js
@@ -17,8 +17,7 @@ const create_country_objects = (countries, store) => Object.entries(countries)
                                                             type: 'country',
                                                             options: [{label: gettext('Individual programs only'), value: 'none'}, ...store.country_role_choices],
                                                             admin_access: store.is_superuser,
-                                                            programs: new Set(country.programs),
-                                                            expanded: true
+                                                            programs: new Set(country.programs)
                                                         }
                                                     }),{})
 
@@ -295,6 +294,7 @@ export default class EditUserPrograms extends React.Component {
         }, () => this.hasUnsavedDataAction())
 
     }
+
 
     clearFilter() {
         const val = ''
--- a/js/pages/tola_management_pages/user/components/edit_user_programs.js
+++ b/js/pages/tola_management_pages/user/components/edit_user_programs.js
@@ -97,7 +97,7 @@ export default class EditUserPrograms extends React.Component {
             filtered_countries: countries,
             filtered_programs: programs,
             ordered_country_ids: store.ordered_country_ids,
-            flattened_programs: flattened_listing(store.ordered_country_ids.filter(id => id in countries).map(id => countries[id]), programs, this.countryStore.isExpanded.bind(this.countryStore)),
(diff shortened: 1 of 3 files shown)


js/pages/tola_management_pages/user/components/edit_user_programs.js
```diff
@@ -33,10 +32,10 @@ const create_program_objects = (programs, store) => Object.entries(programs)
                                                            }),{})
 
 //we need to flatten the country -> program heirarchy to support the virtualized table
-const flattened_listing = (countries, programs) => countries.flatMap(country =>
+const flattened_listing = (countries, programs, isExpanded) => countries.flatMap(country =>
                                                         [
                                                             country,
-                                                            ...(country.expanded ? Array.from(country.programs)
+                                                            ...(isExpanded(country.id) ? Array.from(country.programs)
```
axoplasm, line 38: for my edification: I am not familiar with this three-dot syntax. What is it?

js/pages/tola_management_pages/user/components/edit_user_programs.js
```diff
@@ -33,10 +32,10 @@ const create_program_objects = (programs, store) => Object.entries(programs)
                                                            }),{})
 
 //we need to flatten the country -> program heirarchy to support the virtualized table
-const flattened_listing = (countries, programs) => countries.flatMap(country =>
+const flattened_listing = (countries, programs, isExpanded) => countries.flatMap(country =>
```
axoplasm, line 35: `isExpanded` is a bit opaque to me. Where does this value come from? Is it toggled when I click the expando?

js/pages/tola_management_pages/user/components/edit_user_programs.js
camgnostic: Spread syntax - the return there is an array, so this spreads the elements out into a flat array of [country, program1, program2, etc.] (given the country.programs array of [program1, program2, etc.])

js/pages/tola_management_pages/user/components/edit_user_programs.js
camgnostic: I pass it in whenever it's called, because the specs broke this functional design and required a stateful element to determine if its expanded (a combination of various filter states), so I bound a function to the required states and passed it in each time flattened_listing is called.  Would a @param jsdoc type function docstring help?  Or should I just refactor that whole flattened_listing to be a method on the store (so it has access to internal state)?

axoplasm: a docstring would probably do it, I mean I can tell what it should be telling me (`isExpanded`, duh). I think this + the `...` syntax is kind of confusing to me, like: `flattened_programs` is a 1-dimensional array (of indeterminate length), of n countries, n programs, and the value of `isExpanded`?

Forgive my ignorance here, js just makes my eyes water

camgnostic: >  js just makes my eyes water

you and me both.  Tried adding some jsdoc explaining the paradigm on that flattened listing function and refactored the calls to it so the callback function is more clearly shown (and commented)

PR #85, opened 2020-08-10
1870 log result date

sanjuroj:


js/pages/tola_management_pages/audit_log/__tests__/auditLog.test.js
```diff
@@ -54,4 +55,25 @@ describe("Audit log", () => {
...
+    it("has two components - one for indicator and one for result", () => {
+        const component = renderer.create(
+            <IndicatorNameSpan
+                indicator={auditLogIndicator}
+                result_info={auditLogResult}
+            />
+        ).toJSON();
+        expect(component).toHaveLength(2);
+
+    })
```
camgnostic, line 78: Is it worth either snapshotting this output or testing for the result date info?  Seems like we could be a little more specific in expecting output here to make sure the date/name string is all correct

tola_management/models.py
```diff
@@ -520,6 +520,13 @@ def diff_list_sorter(self, element):
         except ValueError:
             return 999
 
+    @property
+    def result_info(self):
```
camgnostic, line 524: Can we add a comment to this to explain what's using it?  Just because indicators.models is a good demo of how we can load classes up with properties that we won't always remember to remove when we refactor later

js/pages/tola_management_pages/audit_log/__tests__/auditLog.test.js
sanjuroj: Done.  I'm never sure when snapshotting is just going to lead to updating the snapshot without looking at it, and when it's useful.

tola_management/models.py
sanjuroj: Good idea, done.

sanjuroj:
Let's not merge to master

PR #94, opened 2020-08-28
2048 rounding tooltip

camgnostic:
Adds rounding tooltip to:
- results form
- IPTT (web)
- IPTT (excel)

strings marked for translation but translations not supplied yet

### DIFF
--- a/js/pages/iptt_report/components/report/headerCells.js
+++ b/js/pages/iptt_report/components/report/headerCells.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { FaRegQuestionCircle } from 'react-icons/fa';
 
 const HeaderCell = ( props ) => {
     let style = props.styleWidth ? {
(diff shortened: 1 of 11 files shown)


js/pages/iptt_report/components/report/headerCells.js
```diff
@@ -15,6 +16,40 @@ const HeaderCell = ( props ) => {
     )
 }
 
+class LopActualHeaderCell extends React.Component {
+    constructor( props ) {
+        super(props);
+        this.style = props.styleWidth ? {
+            minWidth: `${props.styleWidth}px`
```
sanjuroj, line 23: Do we do a lot of direct styling with our react components?  Would it be better to add css classes for this, just to be consistent with the way to style other react components?

js/pages/iptt_report/components/report/tableHeader.js
```diff
@@ -155,7 +155,7 @@ const ColumnHeaderRow = inject('rootStore')(
                         /* # Translators: Column header for a target value column */
                         gettext('Target')
                     } />
-                <HeaderCells.HeaderCell
+                <HeaderCells.LopActualHeaderCell
                     styleWidth={110}
```
camgnostic, line 159: @sanjuroj  - see what you're saying on the manual styles, but in this section (lines 90-180) you can see that because the layout of the report is very specific column to column I'm setting an explicit width (but _only_ the width) for almost all of the columns.  I could do that with either utility classes (`base-column w-130` or whatever) or with BEM classes (`base-column__unit-of-measure-column`) and set them in SCSS, but that seems like a _less_ DRY solution, given it's adding the class structure to pass just the one property from the stylesheet forward.  I don't like setting HTML style attributes manually, but I think it's less cluttered in this application... what do you think?

PR #102, opened 2020-09-03
2104 qa program

sanjuroj:


camgnostic:
we can probably cut the create_qa_programs_old.py at this point, right?  Looks super solid, love the tests, and it's really easy to follow (and easier to make additions to) - 👍

camgnostic (approved):

sanjuroj:
I thought I might leave the old version in there for just a couple of cycles to make sure we have an alternative if something goes terribly wrong with the new one. I'll delete it after we've had a chance to use the new one a couple of times.

PR #109, opened 2020-09-15
2104 qa program

sanjuroj:


indicators/management/commands/qa_program_widgets/qa_widgets.py
```diff
@@ -137,8 +137,15 @@ def __init__(self, program, country):
         self.sadd_disagg_labels = self.sadd_disagg_obj.disaggregationlabel_set.all()
 
     def create_standard_indicators(self, **kwargs):
-        indicator_ids = self.create_indicators(self.standard_params_base, **kwargs)
-        indicator_ids.extend(self.create_indicators(self.null_supplements_params, **kwargs))
+        try:
```
camgnostic, line 140: Default arg to pop will save you a whole try catch here (`passed_apply_skips = kwargs.pop('apply_skips', None)`)

camgnostic (approved):
one super unimportant syntax shortener, but it all looks good, and the structure of the standard params indicator factory is evolving to be something really cool - might end up refactoring out some of that with the factories models to deduplicate some functionality as a later backstage project?  This looks great.

indicators/management/commands/qa_program_widgets/qa_widgets.py
sanjuroj: Yeah it will

PR #113, opened 2020-09-17
2104 qa program revisions

sanjuroj:
Move permissions updates to the end of the script so they apply to Tolaland programs.

### DIFF
--- a/scripts/mysql.sh
+++ b/scripts/mysql.sh
@@ -83,7 +83,7 @@ then
     echo "backing db $backup"
     #mysqldump --user=$user --password=$password --host=$host $db_name > $backup_path/$db_name-$date.sql
     #mysqldump --user=$user --password --host=$host $db_name > $backup_path/$host-$db_name-$date.sql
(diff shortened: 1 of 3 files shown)


scripts/mysql.sh
```diff
@@ -83,7 +83,7 @@ then
     echo "backing db $backup"
     #mysqldump --user=$user --password=$password --host=$host $db_name > $backup_path/$db_name-$date.sql
     #mysqldump --user=$user --password --host=$host $db_name > $backup_path/$host-$db_name-$date.sql
-    mysqldump --user=$user --password --host=$host $db_name > $backup_path/$host-$db_name-$date.sql
+    mysqldump --no-tablespaces --user=$user --password --host=$host $db_name > $backup_path/$host-$db_name-$date.sql
```
camgnostic, line 86: is this the fix for the PROCESS permission warning on mysqldump?

camgnostic (approved):
Looks great!

scripts/mysql.sh
sanjuroj: Yep

PR #125, opened 2020-10-06
2403 remove notes field

camgnostic:
Remove notes field and corresponding form and validation detritus, copying notes field value to comments field where it exists currently


indicators/models.py
```diff
@@ -1285,8 +1285,6 @@ class Indicator(SafeDeleteModel):
         _("Edit date"), null=True, blank=True, help_text=" "
     )
 
-    notes = models.TextField(_("Notes"), max_length=500, null=True, blank=True)
```
sanjuroj, line 1288: This may have already been addressed in the ticket, but the char length on the notes field is 500, while the length on the comments field is 225.  Maybe we need to increase length of comment field?

sanjuroj (approved):

PR #131, opened 2020-10-16
Gp slow program admin

camgnostic:


tola_management/programadmin.py
```diff
@@ -417,21 +413,115 @@ class ProgramAdminViewSet(viewsets.ModelViewSet):
...
+        """
+        # start with the correctly annotated-for-rf queryset:
+        queryset = Program.rf_aware_objects.all()
+        queryset = queryset.annotate(
+            # this counts users who are not mercy corps users (partner access) who have been assigned:
+            program_access_users_count=models.functions.Coalesce( # coalesce so None is 0 (summable later)
+                models.Subquery(
+                    ProgramAccess.objects.filter(
+                        program=models.OuterRef('pk'),
+                        tolauser__organization_id__gt=1
```
sanjuroj, line 431: Would it make sense to move away from using the hard coded id number of MC's org for these kinds of queries? 
 I assume we've done this all over the app (I'm sure I'm guilty of similar things), but setting up constants somewhere to store this info might be good.  Totally understand if you don't want to do it as part of this patch though.

sanjuroj (approved):
Looks great, just a couple of questions.  Great use of comments.

tola_management/programadmin.py
camgnostic: this is a really good thought, I've been using id != 1 or id = 1 for a while, but that's terrible practice.  I'm gonna file a backstage ticket to look for other instances, but I fixed it in this one

tola_management/programadmin.py
```diff
@@ -417,21 +414,118 @@ class ProgramAdminViewSet(viewsets.ModelViewSet):
...
+                    ).values('users_count')[:1],
+                    output_field=models.IntegerField()
+                ), 0
+            ),
+            # only mercy corps users can be given country-wide access, so count them with filters to minimize joins:
+            country_access_users_count=models.functions.Coalesce( # coalesce so None is 0 (summable later)
+                models.Sum( # sum because there may be multiple countries, each with a user count
+                    models.Subquery(
+                        CountryAccess.objects.filter(
+                            country=models.OuterRef('country'),
```
sanjuroj, line 445: Curious why this isn't `country__in=models.OuterRef('country')`, since Program.country is M2M.

sanjuroj (approved):

tola_management/programadmin.py
camgnostic: I found that the m2m join caused a summing problem - I'm not 100% sure why (m2m joins and annotations have unpredictable effects) so I went with country= and then a Sum, so what it's doing is counting the number of country users _per country_ and then adding them up for all the countries in the program.  That way it gets the right results - I'm not sure if there's a better way to do it?

PR #145, opened 2020-11-17
For mercycorps/TolaActivity #2098, move some state, clear selected pr…

sanjuroj:
…ograms when selected by default is unchecked.


camgnostic (approved):
looks good!  Tests are super clear

PR #152, opened 2020-11-18
For merycorps/TolaActivity#2098, use unified change notice for success.

sanjuroj:


js/pages/tola_management_pages/country/models.js
```diff
@@ -200,16 +200,27 @@ export class CountryStore {
     }
 
     onSaveSuccessHandler({retroProgramCount}={}) {
-        let message = gettext("Successfully saved")
         if (retroProgramCount) {
-            message = interpolate(ngettext(
+            const message = interpolate(ngettext(
```
camgnostic, line 204: maybe commenting in here as to why the save success happens two different ways (pnotify success [which is a top right growler, right?] vs our unified changeset) just to prevent a future dev from thinking that needs to be converted over

camgnostic (changes requested):
just the one comment - otherwise it all makes sense despite git completely failing to identify this diff

sanjuroj:
Closed because it needs to be merged to a different branch.

PR #155, opened 2020-11-19
2098 and 2457 fixes

sanjuroj:
- prevents retro menu from appearing when there are no programs (2098)
- disables form fields while confirmation box is active (2457)


camgnostic (approved):
Nice pivot to using aria already!
Looks great

PR #157, opened 2020-11-19
For merycorps/TolaActivity#2098, make buttons look disabled when they…

sanjuroj:
… are disabled.

### DIFF
--- a/scss/components/_tola_management.scss
+++ b/scss/components/_tola_management.scss
@@ -141,6 +141,7 @@
     }
 }
 
+// formatting the "Assign new disaggregation to all indicators" expando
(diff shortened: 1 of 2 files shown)


scss/components/_tola_management.scss
```diff
@@ -151,6 +152,12 @@
     }
 }
 
+.disagg-form__fieldset:disabled a.btn.btn-link.btn-danger {
```
camgnostic, line 155: TIL you can select on disabled in CSS... this obviates so many helper classes we have elsewhere

camgnostic (approved):

PR #169, opened 2020-12-08
Translations repair

camgnostic:
Pulls in new/old/Marie- translations from the spreadsheet provided.


sanjuroj (approved):
Spot checked updated translations in the .po file with the preferred translations from the spreadsheet and it all looks right.

PR #171, opened 2020-12-09
2457 modal pnotify on country admin

sanjuroj:
- Provide the option to convert PNotify confirmation message into a modal
- Do this for the country admin when retro disaggs are selected and remove the code that manually disabled the other form elements on the page.


js/components/changesetNotice.js
```diff
@@ -67,6 +67,7 @@ const create_unified_changeset_notice = ({
     cancel_text = gettext('Cancel'),
     context = null,
     notice_type = 'notice', // possible values: error (danger/red), info (blue), success (green), notice (warning/yellow)
+    modal = false, // notice will act like a modal with the rest of the page inactive until the notice is responded to
```
camgnostic, line 70: maybe "if set to true, notice will act..."?

js/components/changesetNotice.js
```diff
@@ -191,6 +192,7 @@ const create_unified_changeset_notice = ({
     let cancel_button = {
         text: cancel_text,
         addClass: 'btn-sm',
+        promptTrigger: true,
```
camgnostic, line 195: you've tested this on our text-entry fields and it doesn't result in cancelling when you press enter?  That was my only worry in adding promptTrigger: true on non-modal popups

js/pages/tola_management_pages/country/components/edit_disaggregations.js
```diff
@@ -411,7 +411,7 @@ export class DisaggregationType extends React.Component {
         const managed_data = this.state;
         const retroPrograms = managed_data.id === "new" && Object.values(this.programsForRetro).length > 0 ? <RetroProgramCheckBoxWrapper
                 programs={this.programsForRetro}
-                disabled={this.state.selected_by_default !== true || this.props.formDisabled}
+                disabled={this.state.selected_by_default !== true}
```
camgnostic, line 414: just curious, trying to follow the logic, what is replacing the previous "props.formDisabled"?  Is it because the modal form of the notice blocks the form so we don't need to duplicate by disabling?

camgnostic (approved):
Code's approved.  Had a comment suggestion and a couple questions for my edification only

js/components/changesetNotice.js
sanjuroj: It does but I figured that was better than activating the ok button.  Were you hoping for a more neutral component being activated and the user needing to specifically select Ok or Cancel?

js/pages/tola_management_pages/country/components/edit_disaggregations.js
sanjuroj: Yep, exactly.

js/components/changesetNotice.js
sanjuroj: Yes, that makes sense.

js/components/changesetNotice.js
camgnostic: I'm not sure, I guess this is more a UX question, but my instinct for the most neutral change would be to add `promptTrigger` for non keyboard-input pNotifies, and manually direct focus to the text box in case of input-required ones?  That might take a lot more overhaul.  Maybe this is one to direct to @jennyhavah?

sanjuroj: Oh interesting, I can play around with it a bit and see if it's easy.

PR #182, opened 2020-12-15
date was being compared to UTCNow() on home page and date.now() on pr…

camgnostic:
…ogram page resulting in a window where they disagreed about whether "today" was over.  Moved both to a UTC Now() paradigm (rather than preferencing server location or trying to guess client location for back-end calculations

- not sure about the arrangement, the non-top-level imports seem like a code smell but I'm not sure if it's worth reorganizing to pull utils that wouldn't cause a circular import into their own file or if this is fine or indicating a need for a bigger refactor.  Thoughts?

### DIFF
--- a/indicators/models.py
+++ b/indicators/models.py
@@ -961,11 +961,12 @@ class IndicatorTargetsMixin:
         )
 
     def annotate_most_recent_complete(self):
+        from indicators.queries import utils as query_utils
         return self.annotate(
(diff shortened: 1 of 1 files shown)


sanjuroj:
I don't have a sense of when/how the nested import might cause issues or confusion, so doing this without a refactor for now seems fine.  It does smell a little though.

sanjuroj (approved):

PR #185, opened 2020-12-15
2462 page titles translated

camgnostic:
per mercycorps/TolaActivity#2462 changes page title language to match page header language and translates page titles


sanjuroj (changes requested):
Add translator comments?

PR #195, opened 2020-12-18
2454 sites list program page reactive

camgnostic:
Makes the sites list component of the program page reactive, updating when a site is added to or removed from a result


camgnostic:
Note: in the (temporary) absence of Travis - all tests run (python and node) and passed locally

andrethomas6 (approved):
I checked that there were test written for Python and React logic, strings were marked for translators with comments preceding them, and that there are ARIA attributes. All looks good.

PR #197, opened 2020-12-19
2399 kpi to indicator type

camgnostic:


camgnostic:
Note - dev changes are also included so that I wouldn't have to rewrite the translation-aware sorting algorithm.  The main flow of the changes for this branch are:
- migrate to add the new indicator type, add it to any KPI indicators, and remove the old KPI field
- update the serializers to not call for the old field
- add helper methods so program page (and any other legacy code) can still check whether an indicator is a KPI
- program page styling updates of the KPI badge

indicators/models.py
```diff
@@ -1501,6 +1507,11 @@ def disaggregations(self):
         disaggregations = self.disaggregation.all()
         return self.SEPARATOR.join([x.disaggregation_type for x in disaggregations])
 
+    @property
+    def key_performance_indicator(self):
+        """transition method for key_performance_indicator field removal - may cause unnecessary db calls"""
+        return self.indicator_type.filter(indicator_type="Key Performance Indicator (KPI)").exists()
```
sanjuroj, line 1513: I'm wondering if it would make sense to write a couple of tests that cover places where we're using the hardcoded value of "Key Performance Indicator (KPI)" as a type.  If we change the text we use to designate a KPI, this function would return false, even though it is supposed to be true.

sanjuroj (approved):
I'm wondering if it would make sense to write a couple of tests that cover the hardcoding of the KPI

PR #202, opened 2020-12-21
2452 copy and paste too much text

camgnostic:
closes mercycorps/TolaActivity#2452

Best balance I could strike between boilerplate-for-days and obscure-to-unreadable.  Thoughts?

### DIFF
--- a/templates/indicators/indicator_form_common_js.html
+++ b/templates/indicators/indicator_form_common_js.html
@@ -1028,6 +1028,7 @@
             const $div = $(`#div_id_${fieldName}`);
             let errMsg = null;
             if ($field.val().length > thisLength) {
+                {# Translators: on a form field, as a warning when a user has entered too much text.  %s replaced with a number (e.g. 500) #}
                 errMsg = interpolate("{% trans 'Please enter fewer than %s characters.'|escapejs %}", [thisLength]);
             }
             if (errMsg) {
--- a/templates/indicators/indicator_form_common_js.html
+++ b/templates/indicators/indicator_form_common_js.html
@@ -1014,6 +1014,33 @@
         }
     }
 
+    function getTextAreaLengthValidator(fieldName, length) {
+        function validateField(e) {
+            const thisLength = length;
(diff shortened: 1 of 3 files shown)


templates/indicators/indicator_form_common_js.html
```diff
@@ -1014,6 +1014,39 @@
...
+     */
+    function getTextAreaLengthValidator(fieldName, length) {
+        function validateField(e) {
+            const thisLength = length;
+            const $field = $(`#id_${fieldName}`);
+            const $validation = $(`#validation_id_${fieldName}`);
+            const $div = $(`#div_id_${fieldName}`);
+            let errMsg = null;
+            if ($field.val().length > thisLength) {
+                errMsg = interpolate("{% trans 'Please enter fewer than %s characters.'|escapejs %}", [thisLength]);
```
sanjuroj, line 1031: Add translator comments?

templates/indicators/indicator_form_common_js.html
```diff
@@ -1014,6 +1014,39 @@
         }
     }
 
+    /*
+     * Higher-order function to automate validation internals for a length-validated textarea
+     * Takes the field name (i.e. "justification") and max length (i.e. 400) and returns a validator for
+     * that field, which when called adds invalid classes/markings/labels, and returns false if not valid
+     * called both on the keyup event listeners below, and as part of whole-form validation
+     */
+    function getTextAreaLengthValidator(fieldName, length) {
```
sanjuroj, line 1023: This is a textbook example of a closure.

sanjuroj (changes requested):
Just a missing translator comment to update.  This seems a bit tedious but less bad than any of the other options I can think of.

templates/indicators/indicator_form_common_js.html
camgnostic: and I got there the way I always do - writing the loop without a closure and wondering why everything is validating the last field in the list...

PR #214, opened 2020-12-28
per update on mercycorps/TolaActivity#1500 this removes the "no acces…

camgnostic:
…s" option from dropdowns when the user has user or admin access to the program's country - so the default reverts to "low" for those programs, and so that a user can't select the invalid "no access" option for a user who has access at the country level

### DIFF
--- a/js/pages/tola_management_pages/user/components/edit_user_programs.js
+++ b/js/pages/tola_management_pages/user/components/edit_user_programs.js
@@ -25,15 +25,31 @@ const create_country_objects = (countries, store) => Object.entries(countries)
                                                         }
                                                     }),{})
 
-const create_program_objects = (programs, store) => Object.entries(programs)
-                                                           .reduce((programs, [id, program]) => ({
-                                                               ...programs,
-                                                               [id]: {
-                                                                   ...program,
-                                                                   type: 'program',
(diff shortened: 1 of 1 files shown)


js/pages/tola_management_pages/user/components/edit_user_programs.js
```diff
@@ -25,15 +25,31 @@ const create_country_objects = (countries, store) => Object.entries(countries)
...
+                ...program,
+                type: 'program',
+                options: (programsWithCountryAccess.includes(parseInt(id)) ?
+                            // if given country level access, do not include "no access" choice:
+                            [...store.program_role_choices] :
+                            // otherwise "no access" is a choice:
+                            [{label: NO_ACCESS, value: 'none'}, ...store.program_role_choices]),
+            }
+        }),{});
+}
```
sanjuroj, line 52: All of this commenting is super helpful!

sanjuroj (approved):

js/pages/tola_management_pages/user/components/edit_user_programs.js
camgnostic: this was a really frustrating batch of tickets to work through because I had to relearn the whole weird ad hoc data structure of countries/programs each time, so I'm trying to leave us some breadcrumbs to follow next time!

PR #215, opened 2020-12-28
2452 bugfixes

camgnostic:
bugfixes based on updates to mercycorps/TolaActivity#2452 - specifically text input fields (previous updates covered textareas)
- unit of measure (custom constructed field in indicator.forms - added maxlength feature to constructed widget attrs based on field definition)
- baseline  - added an arbitrary 20 character length (handles up to 10^20 which is larger than any baseline we are likely to encounter)
- target values - added an arbitrary 15 character length (narrower field than baseline)
- target names (events) - added the 255 character length from the model definition

### DIFF
--- a/indicators/forms.py
+++ b/indicators/forms.py
@@ -371,7 +371,9 @@ class IndicatorForm(forms.ModelForm):
         self.fields['name'].required = True
         self.fields['name'].widget = forms.Textarea(attrs={'rows': 3})
         self.fields['unit_of_measure'].required = True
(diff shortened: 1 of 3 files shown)


camgnostic:
Note: we don't currently have test of our javascript/html validations outside of react because we don't have a functioning selenium/geckodriver test setup for non-reactive rendered behavior, and these changes were entirely in that space, so no tests were added

andrethomas6 (approved):

PR #217, opened 2020-12-29
consolidated some unused code (and updated to use the optional chaini…

camgnostic:
…ng operator), added a function to make sure "no access" doesn't show up in a menu where it is inappropriate, even before save (and vice versa - it is in menus where appropriate even before save)

closes mercycorps/TolaActivity#1500 (QA fail, comments indicating updating spec)

- Sanjuro: this still readable?  I feel like I'm pushing the poorly constructed data model here to it's limits (duplicating the options on every program requires a lot of work to be "reactive") - but I think it works for this ticket, so I'd like to avoid the refactor if it isn't necessary.  Up to you if this still feels maintainable?

### DIFF
--- a/js/pages/tola_management_pages/user/components/edit_user_programs.js
+++ b/js/pages/tola_management_pages/user/components/edit_user_programs.js
@@ -259,45 +259,39 @@ export default class EditUserPrograms extends React.Component {
                 return {...country, role: new_val, has_access: false}
             }
         })()
-        /* next update the program role options (if they have country-level access, do not show NO ACCESS option,
-         * but if they do not have country level access, _show_ the NO ACCESS option)
-         * Due to the weird way options are stored (duplicated as a list on every single program object
-         * in the global stateless "programs" variable) we have to modify the state listing from
-         * programs through filtered_programs to flattened_programs to ensure this isn't undone by the next
-         * change.  Refactoring idea: options listing should be a bound method that checks for country
-         * access and dynamically generates options - requires associating country with program (and
-         * dealing with multi-country program edge cases?)
-         */
-        // start with copy of state programs (don't modify state in place):
-        let statePrograms = this.state.programs;
-        this.state.countries[country_id].programs.forEach(programId =>
+        // using the form of setstate that receives the previous state/props as an argument, as we need
+        // to update the programs (based on current state programs list and new country value)
+        this.setState((state, props) => {
+            // this is a reference to the _old_ state programs, so it's safe to modify:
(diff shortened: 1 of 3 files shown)


js/pages/tola_management_pages/user/components/edit_user_programs.js
```diff
@@ -395,37 +424,34 @@ export default class EditUserPrograms extends React.Component {
...
             } else {
                 // program access checkbox:
-                if(this.state.user_program_access.countries[data.country_id] && this.state.user_program_access.countries[data.country_id].has_access) {
+                if (this.state.user_program_access.countries?.[data.country_id]?.has_access) {
+                    // if the user has access to the country level, they have access to the program:
                     return true
                 }
-                return (access.programs[data.id] && access.programs[data.id].has_access) || false
+                // otherwise if the user has access to the program directly:
+                return access.programs?.[data.id]?.has_access || false
```
sanjuroj, line 436: I had not seen this sort of chaining before.  It's nice.

sanjuroj (commented):
I think this is fine.  The data model is making it a bit convoluted, but probably not worth refactoring just for this.  Just the one question about copying the state variable.

camgnostic:
@sanjuroj Re-requesting to have you double-check me - I think this refactor means the "state" I'm accessing and modifying is one I'm supposed to be (a copy of the existing state at the time of re-render) and I'm returning the new state as a copy.  Thoughts?

js/pages/tola_management_pages/user/components/edit_user_programs.js
```diff
@@ -258,17 +259,39 @@ export default class EditUserPrograms extends React.Component {
                 return {...country, role: new_val, has_access: false}
             }
         })()
+        // using the form of setstate that receives the previous state/props as an argument, as we need
+        // to update the programs (based on current state programs list and new country value)
+        this.setState((state, props) => {
+            // this is a reference to the _old_ state programs, so it's safe to modify:
+            let statePrograms = state.programs;
```
sanjuroj, line 266: Any reason not to just use `state.programs` throughout, rather than setting up the reference?

sanjuroj (approved):
That makes a more sense.  Just one comment but it's a nit, not a blocker.

js/pages/tola_management_pages/user/components/edit_user_programs.js
camgnostic: that's why we have code review 🤦  - thanks!

PR #221, opened 2020-12-29
2408 unified pnotify

camgnostic:


templates/indicators/result_form_common_js.html
```diff
@@ -84,7 +84,7 @@
                         // Close modal
                         $('#indicator_results_div').modal('hide');
                         $('#indicator_results_div').trigger('save.tola.result_form');
-                        PNotify.success("Result deleted")
+                        window.unified_success_message("Result deleted");
```
sanjuroj, line 87: Wouldn't hurt to add translator comment, just in case we ask translators to do an overall review at some point.

templates/indicators/result_form_common_js.html
```diff
@@ -378,20 +378,15 @@
...
-                    hide = false;
-                } else {
-                    // URL is fine, message doesn't need additions and should auto-dismiss
-                    noticeType = 'success';
-                    hide = true;
-                }
-                let notice = window.autoDismissingNotification({type: noticeType, message: message, hide: hide, textTrusted: true});
-                if (displayEvidenceWarning) {
+                    let message_text = "{% trans 'However, there may be a problem with the evidence URL.' %}";
+                    message_text += ' <a href="#" id="warning_review">{% trans "Review warning." %}</a>';
```
sanjuroj, line 384: Maybe formulate this as a single string and add a translator comment?

sanjuroj (approved):
Looks good.  The additional translator comments would but nice, but since these are existing strings, not as necessary.

templates/indicators/result_form_common_js.html
camgnostic: I kept it separate to try and keep HTML out of the translator strings because we'd had problems with it - maybe with comments it can be clear?

PR #226, opened 2020-12-31
2491 pnotify in admin

camgnostic:
for mercycorps/TolaActivity#2491

Simple update of styling to admin notifiers - note Jenny requested that admin errors stay in the top right as now (but with updated styling and no longer autodismissing) so that's why they are moved to top right with this update.

### DIFF
--- a/js/pages/tola_management_pages/country/models.js
+++ b/js/pages/tola_management_pages/country/models.js
@@ -211,57 +211,46 @@ export class CountryStore {
                 "Disaggregation saved and automatically selected for all indicators in %s programs.",
                 retroProgramCount
             ), [retroProgramCount])
-            create_unified_changeset_notice({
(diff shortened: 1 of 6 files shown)


js/pages/tola_management_pages/country/models.js
```diff
@@ -344,10 +333,11 @@ export class CountryStore {
                     this.updateHistory(id);
                 }))
             .catch((errors) => {
+                console.log(errors);
```
sanjuroj, line 336: Left in on purpose?

sanjuroj (approved):

js/pages/tola_management_pages/country/models.js
camgnostic: whoops, no - one-line pr coming right up

PR #229, opened 2020-12-31
(not so) big refactor of program roles editor

camgnostic:
for mercycorps/TolaActivity#1500 refactors options to be dynamically generated on the component class, avoiding stale options and also correctly differentiating options for multi-country programs with differing country-level access

### DIFF
--- a/js/pages/tola_management_pages/user/components/edit_user_programs.js
+++ b/js/pages/tola_management_pages/user/components/edit_user_programs.js
@@ -19,34 +19,19 @@ const create_country_objects = (countries, store) => Object.entries(countries)
                                                         [id]: {
                                                             ...country,
                                                             type: 'country',
-                                                            options: [{label: INDIVIDUAL_PROGRAM_ACCESS, value: 'none'}, ...store.country_role_choices],
                                                             admin_access: store.is_superuser,
-                                                            programs: new Set(country.programs)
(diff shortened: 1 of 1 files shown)


js/pages/tola_management_pages/user/components/edit_user_programs.js
```diff
@@ -597,7 +581,7 @@ export default class EditUserPrograms extends React.Component {
                                     id: rowData.id,
                                     disabled: is_role_disabled(rowData),
                                     type: rowData.type,
-                                    options: rowData.options,
+                                    options: this.getRoleOptions.call(this, rowData),
```
sanjuroj, line 584: It would have taken me forever to figure out that I needed to use a `call` method here.  I still don't grok JS on some fundamental level.

sanjuroj (approved):

js/pages/tola_management_pages/user/components/edit_user_programs.js
camgnostic: I definitely got to there by my usual path of not using .call, fighting with the errors, finally thinking to console.log(this), facepalm, add .call.  I use .call and .bind as complete hail marys

sanjuroj: That's my SOP as well.

PR #232, opened 2021-01-05
Add GitHub Action to for Django and JS tests.

sanjuroj:


camgnostic (approved):
This all looks good, and failing python tests are addressed in pull #233 and failing node tests are being addressed by you in another pull, so I think we're good merging this despite fails.  So excited to have travis working again (since we clearly need it given how many test fails piled up while we were travis-less)

PR #237, opened 2021-01-05
Wrapped the folding sidebar contents in a container and added auto ov…

andrethomas6:
- Wrapped the folding sidebar contents in a container
- Added auto overflow property to that container
- Added max height property to that container


camgnostic:
also, those two test failures are addressed in other tickets, so you're clear to merge this one

PR #238, opened 2021-01-06
Fix a couple of front end tests.

sanjuroj:


sanjuroj:
I think once both our PR's are merged in, we'll have passing tests again.

camgnostic (approved):
Thanks for cleaning up the mess I made!

PR #242, opened 2021-01-06
1858 google analytics

camgnostic:
addresses mercycorps/TolaActivity#1858

Also adds:
user type dimension (superuser or no)
user country dimension
site area dimension (defined in a block tag, fall back to "main")
event sends for exports so we can actually see if people are exporting things


sanjuroj (approved):
Looks good!

PR #245, opened 2021-01-12
2453 reorder menu cutoff

andrethomas6:
- Added min width properties to prevent cutting off contents within the component


camgnostic (approved):
I'm happy to know that `max-content` is a thing - somehow I've never seen that before!

PR #253, opened 2021-01-22
1620 iptt visual feedback

andrethomas6:
- Applied shade animation to all of the sidebars select and filter fields. 
- For Selects, will trigger animation when start or end periods change.
- For Filters, will trigger animation in their set methods
- Updated some indentation
- Included the red/teal shine animation and pulse bar animation for future reference, if needed.

### DIFF
--- a/js/pages/iptt_report/components/report/tableBody.js
+++ b/js/pages/iptt_report/components/report/tableBody.js
@@ -6,7 +6,7 @@ import { LevelGroup, IndicatorRow } from './tableRows';
 const ReportTableBody = inject('rootStore', 'filterStore')(
     observer(({ rootStore, filterStore }) => {
         return (
-        <tbody>
+        <tbody className={filterStore._shade ? filterStore._shade : null}>
             {
                 rootStore.levelRows ?
                     rootStore.levelRows.map(
--- a/js/pages/iptt_report/components/report/tableBody.js
+++ b/js/pages/iptt_report/components/report/tableBody.js
@@ -6,7 +6,7 @@ import { LevelGroup, IndicatorRow } from './tableRows';
 const ReportTableBody = inject('rootStore', 'filterStore')(
     observer(({ rootStore, filterStore }) => {
         return (
-        <tbody className={filterStore._shade ? filterStore._shade : null}>
+        <tbody className={filterStore?._shade}>
             {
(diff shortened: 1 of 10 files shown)


js/pages/iptt_report/components/report/tableBody.js
```diff
@@ -6,7 +6,7 @@ import { LevelGroup, IndicatorRow } from './tableRows';
 const ReportTableBody = inject('rootStore', 'filterStore')(
     observer(({ rootStore, filterStore }) => {
         return (
-        <tbody>
+        <tbody className={filterStore._shade ? filterStore._shade : null}>
```
camgnostic, line 9: you don't have to change it, but I just wanna make sure you've seen the new optional chain operator in ES6 - you can write this whole line as `filterStore?._shade` and that does the same thing as the ternary you wrote out (it'll return `undefined` instead of `null` but if that's a problem (shouldn't be, cause react should dump the undefined, but in case) you can do `filterStore?._shade ?? null` and use both the shiny new operators.  Again - not saying you should change it, I'm just proselytizing the optional chain because it's about time we had an operator that saved all that boilerplate

scss/_animations.scss
```diff
@@ -0,0 +1,81 @@
...
+  display: block;
+  top: -50%;
+  left: -50%;
+  z-index: -9;
+  height: 200%;
+  width: 400%;
+  transform: rotate(-135deg);
+  overflow: hidden;
+  background: $gray-700;
+  &-on {
```
camgnostic, line 70: is this how it rotates - like you're taking a block and putting it behind the other one offset right and down then spinning it?  I'm just trying to understand how this works (I'm not super strong in CSS)

camgnostic (approved):

js/pages/iptt_report/components/report/tableBody.js
andrethomas6: Wow, I didn't know that. Thats cool, a shortcut for a shortcut. I made that change, thanks for letting me know.

scss/_animations.scss
andrethomas6: Yeah just about. The block itself is gradient from grey/red/teal/grey and rotated at an angle. Then its sliding down from one corner and behind the other with an offset which gives that border effect.

PR #254, opened 2021-01-23
adds fail mode toggle to ui, fail mode toggle view to urls/views, and…

camgnostic:
… fail mode middleware to intercept and fail requests when active.

Tagging both André and Sanjuro because I want to make sure I'm not adding something potentially disruptive to prod here!

### DIFF
--- a/factories/workflow_models.py
+++ b/factories/workflow_models.py
@@ -116,6 +116,12 @@ class TolaUserFactory(DjangoModelFactory):
     )
     country = SubFactory(CountryFactory)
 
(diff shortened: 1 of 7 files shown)


camgnostic:
I've tested it from a user interaction side pretty intensively, but I'll add automated tests if we like this pattern.  Just wanted to make sure there wasn't something obviously bad about the middleware stack version of this before cementing it with tests

sanjuroj (approved):
This looks awesome!

PR #255, opened 2021-01-23
For mercycorps/TolaActivity#2457, fix modalized PNotify in result form.

sanjuroj:


andrethomas6 (approved):
This change will stop the the existing modal from closing when you cancel out of the PNotify modal, correct?

sanjuroj:
Yep, that's the goal.

PR #258, opened 2021-01-26
1620 iptt visual feedback

andrethomas6:
- Update to not trigger the shade animation until after the page mounts.


js/pages/iptt_report/components/report/tableRows.js
```diff
@@ -202,6 +202,9 @@ class IndicatorRow extends React.Component {
 
     componentDidMount() {
         this.props.rootStore._expandoRows.push(this);
+        setTimeout( () => 
+            this.props.rootStore._mounted = true
+        ,1500)
```
camgnostic, line 207: For an IPTT with, say, 20 indicators, won't it have 20 IndicatorRows and then 20 timeouts waiting to update mounted 20 times?  What's the goal of the setTimeout here?   Is it 1.5s to avoid an initial blink?

camgnostic (commented):
Question on the setTimeout portion below

js/pages/iptt_report/components/report/tableRows.js
camgnostic: this might even be an opportunity to use a hook, the ReportTableBody is a functional component, so it would be easy to drop a useEffect hook in there, which from what I understand will run after all child components have mounted (so it'd be after all the rows load, thus letting you fire it after the whole report is well and truly mounted)?  Caveat: I haven't used useEffect yet, so I'm speaking from having read docs and thinking I understand them, not from experience.

andrethomas6: @camgnostic Good suggestion as it was calling it for every row. I ended up converting the ReportTableBody to a class component because from what I read, hooks can only by used with mobx-react-lite or mobx 6.

camgnostic: oh I didn't even think about the relationship between mobx and hooks - man this ecosystem gets complicated

camgnostic (approved):
good solve!

PR #259, opened 2021-01-26
For mercycorps/TolaActivity#2492, MC org visible but disabled

sanjuroj:
Make MC org visible but disabled for basic admins creating a new user.


camgnostic (approved):
I can see exactly how we go to this state _and_ didn't notice.  Curse you `|| is_superuser`!

PR #261, opened 2021-01-27
For mercycorps/TolaActivity#2518, Edinburgh logins

sanjuroj:
This is to prevent Edinburgh logins from being logged.  

I tried to test the logging by using the log_mock context manager, but it wasn't called.  I'm assuming that's because we don't have logging set up in our test config files.  I didn't feel like this particular instance was a good enough reason to enable it.  

And then I realized that the context managers probably weren't needed since we're not testing the logging and it's not set up, so I removed them.

Let me know if you think I should implement/reimplement any of this stuff.

### DIFF
--- a/tola/pipeline.py
+++ b/tola/pipeline.py
@@ -34,9 +34,10 @@ def create_user_okta(backend, details, user, response, *args, **kwargs):
             country = Country.objects.get(code=attributes["mcCountryCode"])
         except Country.DoesNotExist:
             country = None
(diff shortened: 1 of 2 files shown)


camgnostic (approved):
I agree - I think there's some logging update we could do that's short of "completely overhaul our logging right now from first principles", and when we do that we can add testing to logging in a system-wide way, rather than a one-off for this case - let's talk more about this on Wednesday?

PR #262, opened 2021-01-27
2503 sidebar scrolling iptt

andrethomas6:
- When scroll to the bottom of the IPTT, the sidebar matches it and scrolls along with it.


camgnostic (approved):
Love it!  I wonder if maybe a line or two of comments just to explain what bottomscrolling is doing there would help future us save time tracing things through.  But looks great!

camgnostic (approved):
👍

PR #263, opened 2021-01-28
2511 duplicate pins

camgnostic:
Should prevent pins from saving if they have the same name, program, and user.  Shows an error message to the user if they attempt to.


sanjuroj (approved):
Looks great!

PR #264, opened 2021-01-28
For mercycorps/TolaActivity#2457, change focus to rfc or rationale.

sanjuroj:


camgnostic (approved):
Solid commenting.  Frustrating that the "afterOpen" doesn't mean "after it's actually mounted and has adjusted itself into place"!

sanjuroj:
I know, right?

PR #267, opened 2021-01-28
2503 sidebar scrolling iptt

andrethomas6:
Added a div to extend and offset the footers height. Allows the top of the sidebar and toggle button to remain accessible.


sanjuroj (approved):
This looks good.  My only suggestion is to put a couple of comments in explaining _why_ the extender is necessary.  Like, what problem was occurring that needed this fix.

PR #269, opened 2021-01-28
debug doesn't apply on qa, moved check to get host and reordered to m…

camgnostic:
…inimize every-request overhead


sanjuroj (approved):
Looks good.  I'm curious, have you tried activating fail mode, closing the window, and then reopening?  Can you get to a point where you can turn off fail mode?  It seems like it should be possible, but maybe worth a try if you haven't already.

PR #270, opened 2021-01-28
2511 duplicate pins

andrethomas6:
- If it takes longer that 1.5s to get a response, show the loading spinner for at least 0.5s.
- if response comes in less than 1.5s, go straight to success or fail message.
- Removed the test 'Sending' and styled the spinner.


sanjuroj (approved):
Looks good.

PR #278, opened 2021-02-02
For mercycorps/TolaActivity#2522, adds try/catch to google analytics functions

sanjuroj:


camgnostic (approved):
looks good - I hope this catches things!

sanjuroj:
Problem user was having "fixed itself" after dumping browser data.  Since we couldn't reproduce the problem, we're not sure if this fixes something or might cause other problems.  Closing without merging.

PR #279, opened 2021-02-03
2524 admin logging

sanjuroj:
Enables production server to email error messages if settings are configured properly (DEBUG=False, EMAIL_BACKEND='django.core.mail.backends.smtp.EmailBackend').

### DIFF
--- a/config/sample-settings.secret.yml
+++ b/config/sample-settings.secret.yml
@@ -72,8 +72,6 @@ GOOGLE_ANALYTICS_DOMAIN: 'example.org'
 SECRET_KEY: 'replace with a suitable key'
 
 LOGFILE: '/path/to/log/file'
-# Uncomment this to turn on admin emails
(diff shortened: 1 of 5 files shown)


config/sample-settings.secret.yml
```diff
@@ -72,6 +72,8 @@ GOOGLE_ANALYTICS_DOMAIN: 'example.org'
 SECRET_KEY: 'replace with a suitable key'
 
 LOGFILE: '/path/to/log/file'
+# Uncomment this to turn on admin emails
+# LOGGING_ADMIN_EMAIL_HANDLER_CLASS: django.utils.log.AdminEmailHandler
```
camgnostic, line 76: Do we still need this?

camgnostic (approved):
Approved - but I think there's an extra entry in settings.secret we can take back out

config/sample-settings.secret.yml
sanjuroj: Nope, good catch.

PR #286, opened 2021-02-06
1422 change log

andrethomas6:
- Updated the title in the Change Log export
- Updated translations


tola_management/programadmin.py
```diff
@@ -121,7 +121,7 @@ def _result_disaggregation_serializer(raw_disaggs):
         Cell(ws, value=_('Rationale'))
     ]
 
-    title = Cell(ws, value=_("Change log"))
+    title = Cell(ws, value=_("Program change log"))
```
sanjuroj, line 124: It's probably a good idea to add a translator comment, even though this same string is used in another file that has a translator comment. If you want, just copy the same comment and add it here.

tola_management/programadmin.py
```diff
@@ -608,7 +608,7 @@ def export_audit_log(self, request, pk=None):
         program = Program.objects.get(pk=pk)
         workbook = Workbook()
         workbook.remove(workbook.active)
-        ws = workbook.create_sheet(_('Change log'))
+        ws = workbook.create_sheet(_('Program change log'))
```
sanjuroj, line 611: Translator comment as above.

sanjuroj (changes requested):
Just a couple of translator comments.

tola_management/programadmin.py
andrethomas6: @sanjuroj I added a translator comment. Can you make sure that is the correct syntax?

sanjuroj: Yep, looks good.  If you're unsure, you can always run the makemessages command again and look for the comment in the .po file (either with a plain text editor or the POEdit software).  I sometimes still have to do that with JS files or Django template files, because the syntax can get a bit weird.

sanjuroj (approved):

PR #290, opened 2021-02-08
2500 audit log crash

sanjuroj:
- Audit logs should function with some missing data
- Eliminated "M" imports in tests
- Fixed translation interpolation

@camgnostic I removed what looked like an extra text field in the changelog.js that, as far as I can tell, wasn't doing anything.  Please take a close look at that to double check my logic.


sanjuroj:
@camgnostic Ready to review again.  I deleted the commit that removed the extra title text and added some comments.

factories/tola_management_models.py
```diff
@@ -0,0 +1,149 @@
...
+        model = AuditLogRationaleSelection
+
+    other = Iterator([1,0,0,0,0,0,0,0])
+    adaptive_management = Iterator([0,1,0,0,0,0,0,0])
+    budget_realignment = Iterator([0,0,1,0,0,0,0,0])
+    changes_in_context = Iterator([0,0,0,1,0,0,0,0])
+    costed_extension = Iterator([0,0,0,0,1,0,0,0])
+    covid_19 = Iterator([0,0,0,0,0,1,0,0])
+    donor_requirement = Iterator([0,0,0,0,0,0,1,0])
+    implementation_delays = Iterator([0,0,0,0,0,0,0,1])
```
camgnostic, line 30: this is cool!  Maybe a comment documenting what the Iterator does might be helpful for future maintainers - I imagine it falls out from the weird field that I used for this model, but I could see it being confusing for someone who hadn't written the base model

camgnostic (approved):
one place where a comment might help, but man it is going to be great to not have the random 'M' on all the model names in factories!

factories/tola_management_models.py
sanjuroj: Yep, done.

PR #300, opened 2021-02-18
Add questions

sanjuroj:


indicators/serializers_new/iptt_period_data_serializers.py
```diff
@@ -17,6 +17,8 @@ def __init__(self, period_dict):
             period_dict['count'] = period_dict.get('count', None)
             self.__dict__ = period_dict
 
+    # Is this an override of a parent method?  If not, I can't find its use anywhere.  Is it used?  If not can this
+    # and the PeriodObject above can be deleted maybe?
```
camgnostic, line 21: So this method is actually used in indicators.serializers_new.iptt_indicator_serializers in a couple places (line 347 and 356) - this is a while back but I _think_ the problem was that a serializer needs an object and not just a dict to serialize (is that right?) but sometimes I only had a dict of period data I was working with, so I needed a way to instantiate an object with rough equivalence (I added a couple formatting lines to each from_dict method) and then serialize that object.  I am totally up to rewriting this in a better way, but if not I can just document the daylights out of this (the first step in documenting serializers was I was going to just one-liner docstring all of the classes and methods, then add more comments for specific areas as needed).  What do you think?

sanjuroj: Ah, missed those references.  For what you described, maybe just a comment on why the separate class is necessary?

tola/model_utils.py
```diff
@@ -135,7 +135,7 @@ def get_serializer(*serializer_classes):
                 purpose: (optional) purpose of this mixin (i.e. "RFAware")
                 model (required for exactly one of the composed classes): a django model
                 fields: list of fields corresponding to fields defined on the parent class or base model fields
-                override_fields: boolean - ignore all other mixins' "fields" attributes (deprecarted)
+                override_fields: boolean - ignore all other mixins' "fields" attributes (deprecated)
```
sanjuroj, line 138: I am seeing one use of override_fields in the IPTTQSMixin.  Is this something you were getting away from and didn't quite get rid of all the places where you were using it?

PR #301, opened 2021-02-23
Add comments and questions.

sanjuroj:


camgnostic:
merged this so the mainline will have the test fixes, but as always please reopen with more questions!

PR #302, opened 2021-02-25
2517 add results from iptt

andrethomas6:
- Moved the Gear Icon to the left, next to the View Results Icon.
- Adjusted Icons to vertically align to the top with text.
- Centered align the Change and #/% columns.
- Increased the width of the first column to handle longer level/No. (Ticket #2533)
- Added the Add results button/modal with same functionality as the Program Page.


andrethomas6:
@camgnostic The test that failed here is the one regarding change to Program Change Log causing the test to fail. I think it might be fixed in poblano but not yet in hatchling?

sanjuroj (commented):
Looks good!

PR #304, opened 2021-03-01
Update to adjust expando column width to fit text on one line.

andrethomas6:
Update to adjust expando column width to fit text on one line.

### DIFF
--- a/scss/_iptt_table.scss
+++ b/scss/_iptt_table.scss
@@ -126,10 +126,13 @@ td.iptt-level-row, td.ip-level-row {
 }
 
 td.expando-cell {
-    min-width: 125px;
     color: $blue;
     cursor: pointer;
     border-right: none;
+    div { // Will adjust the width be long enouugh to fit text on one line.
+        display: flex;
+        min-width: max-content;
+    }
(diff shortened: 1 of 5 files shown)


sanjuroj (approved):
Not that you're changing anything related to accessibility, but since you're here, did you check to see if the expando is keyboard accessible?  Do we need to handle Enter commands as well?

andrethomas6:
Hey @sanjuroj I added aria labels, tab and escape key functionality for the IPTT accordion. I also added the ability to escape out of three modals with the escape key.

scss/_iptt_table.scss
```diff
@@ -126,10 +126,13 @@ td.iptt-level-row, td.ip-level-row {
 }
 
 td.expando-cell {
-    min-width: 125px;
     color: $blue;
     cursor: pointer;
     border-right: none;
+    div { // Will adjust the width be long enouugh to fit text on one line.
```
sanjuroj, line 132: typo on "enouugh"

sanjuroj (approved):
Looks good, thanks for making the updates!

PR #308, opened 2021-03-05
For mercycorps/TolaActivity#2517, add read_only flag to view context.

sanjuroj:


sanjuroj:
I added a read_only status to the react_data context.  You should be able to read that whichever component pulls in the data.

andrethomas6 (approved):

PR #312, opened 2021-03-09
2028 rf builder indicator text

andrethomas6:
- Truncate indicator name to fit available space on two lines.
- Split before or after words.
- Add ellipse to the end if truncated.


sanjuroj (approved):
Nice!

PR #315, opened 2021-03-11
Updated formula to calculate the width available for truncated text.

andrethomas6:
Updated the formula to calculate the width available for the indicator name by using the sortable-list-group element's width and subtracting the width of the other elements, margins, and padding.


andrethomas6:
@sanjuroj This is the correct branch to Merge to for next week, right?

PR #316, opened 2021-03-12
For mercycorps/TolaActivity#2555

sanjuroj:
Fix the reporting period modal.  It needs to ignore safe-deleted indicators when figuring out what the latests periodic start date is.

### DIFF
--- a/workflow/views.py
+++ b/workflow/views.py
@@ -17,7 +17,7 @@ from workflow.forms import (
     FilterForm,
 )
 
-from indicators.models import Result
+from indicators.models import Result, Indicator
 
 from django.utils.translation import gettext as _
 from django.views.generic.list import ListView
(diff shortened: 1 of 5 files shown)


workflow/views.py
```diff
@@ -332,6 +332,7 @@ def reportingperiod_update(request, pk):
             pass
         elif (program.last_time_aware_indicator_start_date and
               reporting_period_end.date() < program.last_time_aware_indicator_start_date):
+            print('last start date', program.last_time_aware_indicator_start_date)
```
andrethomas6, line 335: Did you want to keep this print statement or is it left over from your work?

andrethomas6 (approved):

workflow/views.py
sanjuroj: Ah, good catch, thanks!

PR #321, opened 2021-03-22
2401 freq models

sanjuroj:
@andrethomas6 Turns out there was little to no real front-end work to do for this ticket, so there's not thing you will need to code for it.

- Updates reporting frequency and data collection frequency models to make then many-to-many with Indicators and to add a sort order field.  Also specifies that the frequency value must be unique.
- Updates Indicator form to accommodate new model
- Updates IPTT display to list a comma delimited list of frequencies for both fields.


indicators/migrations/0096_reporting_and_data_collection_updates.py
```diff
@@ -0,0 +1,100 @@
...
+            name='sort_order',
+            field=models.IntegerField(null=True),
+            preserve_default=False,
+        ),
+        migrations.RunPython(insert_sort_order, reverse_code=migrations.RunPython.noop),
+        migrations.AddField(
+            model_name='indicator',
+            name='data_collection_frequencies',
+            field=models.ManyToManyField(
+                help_text='How frequently will you collect data for this indicator? The frequency and timing of data collection should be based on how often data are needed for management purposes, the cost of data collection, and the pace of change anticipated. If an indicator requires multiple data sources collected at varying frequencies, then it is recommended to select the frequency at which all data will be collected for calculation.',
```
andrethomas6, line 56: Shouldn't this read, "The frequency and timing of data collection should be based on how often data '***is***' needed for management purposes,..."

andrethomas6 (approved):

indicators/migrations/0096_reporting_and_data_collection_updates.py
sanjuroj: Hmmm, yes? I'll run it by Jenny. The existing text wasn't really changed as part of this ticket, but if it's wrong we should fix it.

PR #329, opened 2021-03-25
Update to show bulk import banner alert only after the first level wa…

andrethomas6:
- Update to show bulk import banner alert only after the first level was saved. 
- Changed hiding banner flag from session storage to local storage to work across tabs.


andrethomas6:
@sanjuroj I made the changes for storing the import banner's show/hide flag for a user's session as we discussed.

- When a user closes the banner, a flag is saved in the browsers local storage and also an ajax request is sent to save the flag in Django's session storage.
- When the RF page mounts, it will check the browsers local storage for the flag and also send an ajax request to get the flag from Django session storage. 
- The Banner will not show if the flags says to hide or not display it in either of these locations.
- Having it stored and retrieved from both locations allows the functionality to still work if theres a failure in saving or retrieving from the local storage or Django session storage.

In the _views.py_, let me know if you see anything that should changed for handling the GET request in python. The PUT request was already written for this url so I just added the GET case to it.

tola/views.py
```diff
@@ -201,6 +201,18 @@ def update_user_session(request):
...
+        try:
+            query = request.GET["query"]
+            dump = json.dumps({"result": request.session.get(query)})
+            return HttpResponse(dump, content_type='application/json', status=200)
+        except:
+            error = "Error getting session variables (request params {0})".format(query)
+        if error is not None:
+            logger.error(error)
+            return HttpResponse(error, status=500)
+        return HttpResponse(status=204)
```
sanjuroj, line 215: Could you just get rid of the `if error is not None` and handle the logging and error return in the `except` block?  Then you probably wouldn't need the default `return HttpResponse(status=204)`?

js/apiv2.js
```diff
@@ -91,8 +91,27 @@ const api = {
...
+        return await axios.get('/update_user_session/', {
+            params: {
+                query: query
+            }
+        })
+        .then(response => response.data)
+        .catch(this.logFailure)
+    },
+    updateSessions (sessionVarsToUpdate) {
+        axios.put('/update_user_session/', sessionVarsToUpdate, {
```
sanjuroj, line 105: Would it make sense to put the axios call into a form similar to ones at the top of the file and then call it the way the ones above are called?

sanjuroj (approved):

sanjuroj (approved):

PR #342, opened 2021-04-07
for mercycorps/TolaActivity#2047

sanjuroj:
- Update temporary translations to have diacritics.
- Update sorting of strings without diacritics.


andrethomas6 (approved):
Does the diacritics have a special meaning here or is it just a prefix for the translations?

sanjuroj:
@andrethomas6 There's no special meaning.  We're adding them to ensure that the sorting is done right, if say a translated word contains one.  Unless sorted correctly, "Äpple" will come after after "Orange".  Adding them to the temp translations means we can confirm that the lists of menu items are being sorted the way we want them to.  Also, it's not a bad thing to display some characters with diacritics just to catch any badly handled encoding (which is much less likely to happen in Python3).

PR #344, opened 2021-04-09
2567 password reset

sanjuroj:
- Fix broken password reset button
- Dismiss spinner on error
- Change success Pnotify to failure Pnotify on failure
- Update Tola Activity to TolaData


andrethomas6 (approved):
Good Catch, I don't know why I didn't think to search of Tola **SPACE** Activity.

sanjuroj:
Just shows how bad we are at branding :-)

PR #350, opened 2021-04-14
2536 bulk import popover rf

andrethomas6:
- Created bulk import popover button with the basic option (will take care of the excel download and upload functionality in a different ticket)


js/apiv2.js
```diff
@@ -109,6 +116,22 @@ const api = {
...
             .then(response => response.statusText)
             .catch(this.logFailure)
+    },
+    async downloadTemplate () {
+        // TODO
+            // return await this.templatesInstance.get('/')
+        console.log("API request to get Templates");
+        return await Promise.resolve({data: "EXCEL TEMPLATE"})
+            .then(response => response.data)
+            .catch(this.logFailure);
```
sanjuroj, line 126: Are we going to provide user feedback in some way if the api call fails?  Not sure if that is supposed to be part of this ticket or not, or if it's been speced out at all.

js/apiv2.js
```diff
@@ -109,6 +116,22 @@ const api = {
...
+    },
+    async uploadTemplate(data) {
+        // TODO
+            // return this.templatesInstance.put('/', data)
+        console.log("API request to send Templates");
+        return await Promise.resolve( {statusText: "OK"} )
+            .then(response => response.statusText)
+            .catch(this.logFailure)
```
sanjuroj, line 134: Same question about failed api calls here.

js/components/helpPopover.js
```diff
@@ -34,6 +34,7 @@ export default class HelpPopover extends React.Component {
 
 export class BootstrapPopoverButton extends React.Component {
     popoverName = 'base';
+    popoverTitle = '';
```
sanjuroj, line 37: How does popoverTitle get updated?  Should it be set to accept a props value?

js/components/ImportIndicatorsPopover.js
```diff
@@ -0,0 +1,122 @@
...
+                {(() => {
+                    switch(this.state.status) {
+                        case this.ONE:
+                            return (
+                                <div className="importIndicators-body">
+                                    <div className="import-text">
+                                        <ol>
+                                            <li>
+                                                {
+                                                    // # Translators: TODO
```
sanjuroj, line 44: Did you mean to do these translator comments?

js/components/ImportIndicatorsPopover.js
```diff
@@ -0,0 +1,122 @@
...
+                                        >
+                                            {
+                                                // # Translators: TODO
+                                                gettext("Upload template")
+                                            }
+                                        </button>
+                                    </div>
+                                </div> 
+                            );
+                        case this.TWO:
```
sanjuroj, line 81: Is this a placeholder for something?

js/pages/results_framework/components/level_list.js
```diff
@@ -60,11 +61,16 @@ class LevelList extends React.Component {
...
+    NEW = 1;
+    FIRST = 2;
+    EXISTING = 3;
     constructor(props) {
         super(props);
         this.state = {
-            show_import_banner: null
+            show_import_banner: null,
+            levelsString: "",
+            level_status: this.EMPTY,
```
sanjuroj, line 73: A couple of comments would be helpful to indicate what these new state variables do.  They're not as self-explanatory as "show_import_bannner". Oh, and now that we've had a bit of a conversation about what you need to send to the backend, is some of this level tracking here not necessary?

sanjuroj (changes requested):
One general question, are you planning to generate the new .po files as part of this PR?  Or are you holding off until later.  If you're planning to do it now, you should probably make sure that you're rebased/merged on top of the latest hatchling, or else you'll probably get merge conflicts (you may have already done this).

js/apiv2.js
andrethomas6: Yeah I plan on updating/handling the api calls on the Generate Excel ticket #2543. We can alert users either by a view in the popover or with PNotify for api failed try again.

js/components/helpPopover.js
andrethomas6: This is the base class used for multiple popovers. None of the other instances we already used have a title, like the pin popover. So I just added this option so that in the Import Indicator popover instance I could specify a title while keeping all other instance title free. You can see it defined in the ImportIndicatorsPopover.js file in the ImportIndicatorsButton class.

js/components/ImportIndicatorsPopover.js
andrethomas6: I added these translator comments. Also, I had merge hatchling a few commits back but just merged hatchling again today just in case. I didn't get any conflicts.

js/components/ImportIndicatorsPopover.js
andrethomas6: Yes, this was a placeholder for the other views of the popover. I updated this in the latest commit to give a descriptive name for these views: Download_Upload, Feedback, Confirm, and Success views.

js/pages/results_framework/components/level_list.js
andrethomas6: I added some comments for these status states. This is mainly for handling what buttons show above the main RF Builder panel component when first starting out building the RF Framework.

1.  Before a RF Level has been saved - Show no buttons ("Empty" or "New" status)
2. One RF Level saved - show the import indicators button and banner alert ("First" status)
3. More than one level saved - show the expand, collapse, import indicators, banner alert, and excel buttons ("Existing" status)

js/apiv2.js
andrethomas6: Same as above

sanjuroj (approved):
Looks good!

PR #352, opened 2021-04-23
2536 bulk import popover rf

andrethomas6:


andrethomas6:
Minor styling updates and created temp translations

PR #354, opened 2021-04-23
2546 advanced indicators import

andrethomas6:
-  Converted ImportIndicatorPopover to functional component
-  Used React Context for Popover state management
-  Added the advanced option to the Popover
-  Updated the styling of the Popover
-  Slight update to the Download and Upload methods but will handle most of that on a different ticket
-  Ran temp translations


js/components/ImportIndicatorsPopover.js
```diff
@@ -1,142 +1,271 @@
...
+    popoverName = "importIndicators"
+    popoverTitle = "Import indicators"
 
     constructor(props) {
         super(props);
         this.state = {
-            status: this.DOWNLOAD_UPLOAD
+            levelStore: this.props.levelStore, 
+            chosenTier: this.props.levelStore.tierTemplates[this.props.levelStore.chosenTierSetKey].tiers, // State containing a list of tier levels names of the chosen RF framework tier template
+            tierLevelsRows: [], // State to hold the tier levels name and the desired number of rows for the excel template
```
sanjuroj, line 17: Would it make sense to keep the state of tierLevelsRows and chosenTier in the AdvancedImport component (or the ordinary popup component, depending on where you bind the handlers) rather than the button component, since it's not used in the button component?

js/components/ImportIndicatorsPopover.js
```diff
@@ -1,142 +1,271 @@
...
     }
 
-    handleDownload = () => {
+    setTierLevelsRows = (i, tier, value) => {
+        let updatedTiers = this.state.tierLevelsRows;
+        updatedTiers[i] = {[tier] : value};
+        this.setState({ tierLevelsRows: updatedTiers });
+    }
+
+    getPopoverContent = () => {
```
sanjuroj, line 28: Maybe just add a comment that this is overriding a method in the BootstrapPopoverButton.

js/components/ImportIndicatorsPopover.js
```diff
@@ -1,142 +1,271 @@
...
+
+export class ImportIndicatorsButton extends BootstrapPopoverButton {
+    popoverName = "importIndicators"
+    popoverTitle = "Import indicators"
 
     constructor(props) {
         super(props);
         this.state = {
-            status: this.DOWNLOAD_UPLOAD
+            levelStore: this.props.levelStore,
```
sanjuroj, line 15: As we discussed, pass this directly into Context rather than running it through state.

js/components/ImportIndicatorsPopover.js
```diff
@@ -1,142 +1,271 @@
...
-                                    </div>
-
-                                    <div className="import-buttons">
-                                        <button
-                                            type="button"
-                                            className="btn btn-sm btn-primary btn-download"
-                                            onClick={() => this.handleDownload()}
-                                            >
+    // Creates the default number of rows for the Tier Levels
+    chosenTier.map((tier, i) => {
```
sanjuroj, line 103: You could turn this into a function that is used to create the defaults for the state when the state is created, or something similar that more tightly couples the defaults with state creation.

sanjuroj (changes requested):

andrethomas6:
@sanjuroj I am finished updating this PR and its ready for you to review.

js/components/ImportIndicatorsPopover.js
```diff
@@ -1,157 +1,273 @@
...
+                        gettext('By default, the template will include 10 or 20 indicator rows per result level. Adjust the numbers if you need more or fewer rows.')
+                    }
+                </p>
+                {tierLevelsRows.map((level, i) => <AdvancedOptions key={i} level={level} i={i} />)}
+            </div>
+        </div>
+    )
+}
+
+const AdvancedOptions = ({ level, i }) => {
```
sanjuroj, line 244: Ok, once again, super nitpicky/stylistic, feel free to ignore.  When I see the name AdvancedOptions, I think it's going to be a section of various different toggles.  But it's really just a row of similar dropdown fields.  So maybe just rename to something like IndicatorCountOption, or something like that?  The singular would also make it clear that it's a single row of many.

sanjuroj (changes requested):
Mostly just nitpicky/stylistic (i.e. ignorable if desired) comments on names.  But there may still be one outstanding issue of copy of state.

js/components/ImportIndicatorsPopover.js
andrethomas6: I renamed this component to LevelIndicatorCount and updated the classes to match the same logic.

js/components/ImportIndicatorsPopover.js
```diff
@@ -1,157 +1,273 @@
...
+    const { tierLevelsRows, setTierLevelsRows } = useContext(ImportIndicatorsContext);
+
+    let options = [0, 5, 10, 15, 20, 25];
+    let levelName = Object.keys(level)[0];
+    const [levelValue, setLevelValue] = useState(Object.values(level)[0]);
+
+    let handleSelect = (event) => {
+        setLevelValue(parseInt(event.target.value));
+        let updatedTiers = [...tierLevelsRows];
```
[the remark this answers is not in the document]
andrethomas6: I created a new object and assigned it to the array. How does that look?

js/components/ImportIndicatorsPopover.js
```diff
@@ -1,157 +1,273 @@
...
-            status: this.DOWNLOAD_UPLOAD
-        }
     }
 
-    handleDownload = () => {
-        api.downloadTemplate(this.props.program_id)
+    // Overwritting a method in the BootstrapPopoverButton and provides the content for when the Import indicators button is clicked
+    getPopoverContent = () => {
+        let program_id = this.props.levelStore.program_id; 
+        let chosenTier = [...this.props.levelStore.tierTemplates[this.props.levelStore.chosenTierSetKey].tiers];
```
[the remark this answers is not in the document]
andrethomas6: Makes sense, I've made the update.

js/components/ImportIndicatorsPopover.js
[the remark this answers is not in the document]
sanjuroj: In this case, you're returning some of the original objects (unmodified), along with a new object.  I honestly have no idea if this is considered an anti-pattern or not.  If you think it's not, they go ahead and run with it.  

If you want to make sure you're starting with all fresh values, you could do something like `updatedTiers - tierLevelRows.map((o)=>{return {...o}})`.  But like I said, if you think it's not a problem to use a mixed array, then go with what you've got.

sanjuroj (approved):
One follow-on comment related to copying state.  But in general looks good.

PR #357, opened 2021-04-27
Updated API call to download template and handle fails.

andrethomas6:
- Updated API call to download template 
- Added handling API fails and change to a temporary error view

### DIFF
--- a/js/apiv2.js
+++ b/js/apiv2.js
@@ -37,7 +37,7 @@ const api = {
     }),
     templatesInstance: axios.create({
         withCredentials: true,
-        responseType: 'json',
+        responseType: 'blob',
         headers: {
             "X-CSRFToken": document.cookie.replace(/(?:(?:^|.*;\s*)csrftoken\s*\=\s*([^;]*).*$)|^.*$/, "$1")
         }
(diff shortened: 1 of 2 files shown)


js/apiv2.js
```diff
@@ -117,13 +117,21 @@ const api = {
...
-        // TODO
-            // return await this.templatesInstance.get('/')
-        console.log("API request to get Templates");
-        return await Promise.resolve({data: "EXCEL TEMPLATE"})
-            .then(response => response.data)
-            .catch(this.logFailure);
+    async downloadTemplate (program_id) {
+            return await this.templatesInstance.get(`/indicators/bulk_import_indicators/${program_id}/`)
+            .then(response => {
+                const url = window.URL.createObjectURL(new Blob([response.data]));
```
sanjuroj, line 123: Nice, treating as a blob is what I was missing when I tried the link create method.

sanjuroj (approved):

js/apiv2.js
andrethomas6: Cool, we were thinking along the same lines then.

PR #360, opened 2021-05-03
2546 advanced indicators import

andrethomas6:
- Added a "used" variable to track what RF levels have been created/used.
- Restructured the Tier Level object(params) sent to the backend and added the used variable.
Screen Shot 2021-05-03 at 10 34 29 AM

- Updated the dropdowns to the Select component with customs style to reduce their height.
- Utilized jquery $.extend to create deep copies.

### DIFF
--- a/js/components/ImportIndicatorsPopover.js
+++ b/js/components/ImportIndicatorsPopover.js
@@ -17,7 +17,17 @@ export class ImportIndicatorsButton extends BootstrapPopoverButton {
     // Overriding a method in the BootstrapPopoverButton and provides the content for when the Import indicators button is clicked
     getPopoverContent = () => {
         let program_id = this.props.levelStore.program_id; 
-        let chosenTiers = [...this.props.levelStore.tierTemplates[this.props.levelStore.chosenTierSetKey].tiers];
+        let chosenTierTemplate = [...this.props.levelStore.tierTemplates[this.props.levelStore.chosenTierSetKey].tiers];
+        let chosenTiers = [];
+        chosenTierTemplate.map((tier, i) => {
+            chosenTiers[i] = {
+                name: tier,
+                used: false,
+            }
+        })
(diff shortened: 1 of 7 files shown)


js/components/ImportIndicatorsPopover.js
```diff
@@ -16,10 +17,18 @@ export class ImportIndicatorsButton extends BootstrapPopoverButton {
 
     // Overriding a method in the BootstrapPopoverButton and provides the content for when the Import indicators button is clicked
     getPopoverContent = () => {
-        let program_id = this.props.levelStore.program_id; 
-        let chosenTiers = [...this.props.levelStore.tierTemplates[this.props.levelStore.chosenTierSetKey].tiers];
+        let chosenTiers = [];
+        this.props.chosenTierTemplate.map((tier, i) => {
```
sanjuroj, line 21: This is a nit but it might avoid some potential confusion.  `chosenTierTemplate` in the rest of the RF is referring to a single value (e.g. "Mercy Corps", "European Commission (EC)", etc...  `chosenTiers` represents the list of tiers that belong to that template.  So you might want to rename `chosenTierTemmplate` to something like chosenTiers and chosenTiers to soemthing like `annotatedTiers`, or whatever you feel like naming it.  This would be more consistent with the terminology being used elsewhere in the application.

sanjuroj (changes requested):

js/components/ImportIndicatorsPopover.js
andrethomas6: 1. Updated `chosenTierTemplate` to `chosenTiers` 
2. Updated `chosenTiers` to `tierLevelsUsed` so now it matches `tierLevelsRows`.

- `tierLevelsUsed` object has keys **_name_** and **_used_** 
- `tierLevelsRows` object has keys **_name_** and **_rows_**

PR #366, opened 2021-05-07
Fix byte character mark appearing in Logframe Excel download

sanjuroj:
Removing unicode processing from Logframe views fixes the byte character mark appearing in the Excel download.


sanjuroj:
@andrethomas6 This PR is for mercycorps/TolaActivity#2579.

andrethomas6 (approved):

PR #367, opened 2021-05-10
2547 import complete happy path

andrethomas6:
- Added the Confirm and Success View (added temporary Feedback, Error and Loading views but will complete on another ticket)
- Added loading spinner logic for api calls
- Update API call for upload but to mock an api response. Will need to update with real api calls when available.
- Updated classnames


js/apiv2.js
```diff
@@ -130,16 +130,42 @@ const api = {
...
     },
     async uploadTemplate(data) {
         // TODO
-            // return this.templatesInstance.put('/', data)
+            // Send to backend
         console.log("API request to send Templates");
+            return await Promise.resolve( {statusText: "OK", data: {valid: 16, invalid: 0}} )
+                .then(response => new Promise( resolve => {
+                    let timeOptions = [500, 900, 1000, 2000, 3000]
+                    let delay = timeOptions[Math.floor(Math.random() * 5)]
```
sanjuroj, line 143: Add a comment about what this is for?

sanjuroj (approved):

js/apiv2.js
andrethomas6: I added some comments. This is temporary code to mock api responses at different time delays to see how the loading spinner handles it. I'll replace all this with the actually call when its ready.

PR #372, opened 2021-05-14
2538 bulk import program page

andrethomas6:
- Added the levels and levelTiers to the Program page context.
- Added the import indicators popover on the Program Page.
- Added conditionals based on which page the popover is opened from.

### DIFF
--- a/js/pages/program_page/components/indicator_list.js
+++ b/js/pages/program_page/components/indicator_list.js
@@ -287,8 +287,7 @@ export class IndicatorListTable extends React.Component {
 
 
 const IndicatorListTableButtons = observer(function ({program, rootStore, ...props}) {
-    let chosenTiers = [];
-    rootStore.levelTiers.map((level, i) => chosenTiers[i] = level.name ); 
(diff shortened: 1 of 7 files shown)


js/pages/program_page/components/indicator_list.js
```diff
@@ -286,17 +287,27 @@ export class IndicatorListTable extends React.Component {
 
 
 const IndicatorListTableButtons = observer(function ({program, rootStore, ...props}) {
+    let chosenTiers = [];
+    rootStore.levelTiers.map((level, i) => chosenTiers[i] = level.name );
```
sanjuroj, line 291: Can you consolidate these lines so it's like this?
`let chosenTiers = rootStore.levelTiers.map((level, i) => chosenTiers[i] = level.name ); `

sanjuroj (approved):
Just one optional tweak.

js/pages/program_page/components/indicator_list.js
andrethomas6: Cool, I made that change.

PR #377, opened 2021-05-19
2547 happy import

sanjuroj:
Sorry André, this PR is kind of large.  I modified some JS files that you are probably working on, in order to wire up the frontend and backends.  I mostly just commented your code if I needed to change it, in case I made a hash of it and you need to go back some form of the original.  

I think all the backend endpoints are now there, though the functionality isn't close to being fully developed. But getting the backend and frontend talking will be nice.


indicators/views/bulk_indicator_import_views.py
```diff
@@ -0,0 +1,425 @@
...
+
+            file_name = f'{datetime.strftime(datetime.now(), "%Y%m%d-%H%M%S")}-{request.user.id}-{program.pk}.json'
+            file_obj = BulkIndicatorImportFile.objects.create(
+                user=request.user.tola_user,
+                program=program,
+                file_name=file_name,
+                file_type=BulkIndicatorImportFile.INDICATOR_DATA_TYPE)
+            with open(file_obj.file_path, 'w') as fh:
+                fh.write(json.dumps(new_indicators_data))
+            return JsonResponse({'message': 'success', 'valid': len(new_indicators_data), 'invalid': 0}, status=200)
```
andrethomas6, line 383: Making sure I understand this correctly, we are keeping at most 1 redline feedback file per user id per program and when the request comes in for the redline feedback file, you are using the user id and program to get it? So we don't need to send/receive a user id to/from the front end.

andrethomas6 (approved):
Nice, looks good! I figured the api file was going to need updates once we connect the front and back ends together.

indicators/views/bulk_indicator_import_views.py
sanjuroj: Correct about the just keeping one file per user per program.  The user id is inherently part of the request.  It is extracted by, I believe, the Django session middleware and provided as `request.user` to the Django view.  So we don't have to include it explicitly in the request (and indeed we wouldn't take a user id submitted in that way at face value since anyone could formulate a request with a someone else's user id).

PR #387, opened 2021-06-09
Added a class to the import popover and used it as a selector for sty…

andrethomas6:
- Added a class to the import popover and used it as a selector for styling. 
- Removed styling for default popovers to prevent effecting other popovers.
- Removed styling made previously to overwrite default popovers styling.


sanjuroj (changes requested):
camelCase is a bit unusual for class names.  Let's use BEMish naming conventions as much as possible.  Maybe something like `indicator-import__popover`?

andrethomas6:
@sanjuroj  I updated all the class names to be more BEMish.

andrethomas6:
@sanjuroj I updated the class names again.

PR #391, opened 2021-06-16
Update custom tiers validation

andrethomas6:
Added custom tiers validation for colons


sanjuroj (approved):
Looks good.  Once you've merged, go ahead and assign the ticket to me.  I should do the validation on the backend as well, just in case something escapes the front end somehow.

PR #392, opened 2021-06-16
For mercycorps/TolaActivity#2590, 

sanjuroj:
Create backend validation for having no colons or commas in tier names.

André note that the diff for views_results_framework is really bad.

### DIFF
--- a/indicators/tests/test_rf_views.py
+++ b/indicators/tests/test_rf_views.py
@@ -40,12 +40,19 @@ class TestSaveCustomTemplateView(test.TestCase):
 
     def test_illegal_chars(self):
         self.client.force_login(self.tola_user.user)
+
+        # Test comma
         response = self.client.post(
             reverse('save_custom_template'),
             {'program_id': self.program.id, 'tiers': ['this, ', ' that']},
             content_type="application/json")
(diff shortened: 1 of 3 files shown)


indicators/tests/test_rf_views.py
```diff
@@ -38,6 +38,15 @@ def test_template_saved(self):
         # Note that the extra white space on either side of the tier should be trimmed before saving.
         self.assertEqual(LevelTierTemplate.objects.get(program=self.program).names, ['this', 'tha t'])
 
+    def test_illegal_chars(self):
+        self.client.force_login(self.tola_user.user)
+        response = self.client.post(
+            reverse('save_custom_template'),
+            {'program_id': self.program.id, 'tiers': ['this, ', ' that']},
```
andrethomas6, line 45: Did you want to test if theres colons as well or just commas is fine for an illegal character?

andrethomas6 (approved):

indicators/tests/test_rf_views.py
sanjuroj: Well that would make sense since that's the point of the ticket.

PR #394, opened 2021-06-17
For mercycorps/TolaActivity#2554, create additional change log machin…

sanjuroj:
…ery, add creation of log items to bulk import views, fix baseline value saving, add tests.


andrethomas6 (approved):
Looks good!

PR #399, opened 2021-06-25
Update error codes and graceful handling

andrethomas6:


js/apiv2.js
```diff
@@ -139,24 +139,23 @@ const api = {
...
+        //         let delay = timeOptions[Math.floor(Math.random() * 5)]
+        //         setTimeout(() => {
+        //             resolve( response )
+        //         }, delay);
+        //     }))
+        //     .catch((error) => {
+        //         this.logFailure(error)
+        //         return {error};
+        //     })
+
```
sanjuroj, line 158: Would it make sense to put this in some unit tests? Or maybe you've already done that.

sanjuroj (approved):
Looks good.  Just the one comment about the commented out testing code.

js/apiv2.js
andrethomas6: Yup, I have another branch that I'm working on testing and will include tests for the different response scenarios.

PR #401, opened 2021-06-25
For mercycorps/TolaActivity#2539, update download file name and trans…

sanjuroj:
…late file name and tab name.


indicators/views/bulk_indicator_import_views.py
```diff
@@ -451,7 +452,7 @@ def post(self, request, *args, **kwargs):
         ProgramAuditLog.log_template_uploaded(request.user, program)
 
         wb = openpyxl.load_workbook(request.FILES['file'])
-        ws = wb.get_sheet_by_name(TEMPLATE_SHEET_NAME)
+        ws = wb.worksheets[0]
```
andrethomas6, line 455: Just thinking of an edge case, if a user for some reason adds other sheets to the template and uploads it, would that cause a problem?

andrethomas6 (approved):

indicators/views/bulk_indicator_import_views.py
sanjuroj: It would probably fail when trying to read the program, which I think is ok.

sanjuroj: I've been trying to avoid having to do reverse translations, which I think are doable, but a little weird.

PR #413, opened 2021-07-15
2551 sad path

sanjuroj:


andrethomas6 (approved):
Nice, you got the warning symbol in there.

PR #433, opened 2021-08-02
2537 alert bar bulk import

sanjuroj:


indicators/models.py
```diff
@@ -1066,6 +1064,13 @@ class Indicator(SafeDeleteModel):
         (EVENT, _('Event'))
     )
 
+    FORM = 1
+    IMPORT = 2
+    CREATED_BY_TYPES = (
+        (FORM, _('Form')),
+        (IMPORT, _('Import')),
+    )
+
```
andrethomas6, line 1073: This is left over from my work so we can delete it out.

andrethomas6 (approved):

PR #483, opened 2021-12-28
For mercycorps/TolaActivity#2650, update participant count script.

sanjuroj:


indicators/management/commands/create_participant_count_indicators.py
```diff
@@ -1,230 +1,194 @@
...
+                    'selected_by_default': True,
+                }
+            )
+            if created:
+                for i, label in enumerate(sectors_list):
+                    DisaggregationLabel.objects.create(
+                        label=label, disaggregation_type=disagg_sector_without_double, customsort=i + 1)
+                print('Created disaggregation type: Sectors Direct without double counting')
+            else:
+                print('This disaggregation type already existed: Sectors Direct without double counting')
```
andrethomas6, line 129: Shouldn't this disaggregation be `Sectors Indirect with double counting`?

indicators/management/commands/create_participant_count_indicators.py
```diff
@@ -1,230 +1,194 @@
...
+                    'selected_by_default': True,
+                }
+            )
+            if created:
+                for i, label in enumerate(actual_disagg_labels):
+                    DisaggregationLabel.objects.create(
+                        label=label, disaggregation_type=disagg_actual_without_double, customsort=i + 1)
+                print('Created disaggregation type: Actual without double counting')
+            else:
+                print('This disaggregation type already existed: Actual without double counting')
```
andrethomas6, line 164: I see actuals with and without double counting but do we need to also specify indirect and direct here?

andrethomas6 (approved):

PR #492, opened 2022-01-12
2647 create pc results form

andrethomas6:
Here are the updates for the new PC Results Form so far.
- PC Results Form (Added to Program Page and IPTT)
- API Updates
- Admin type addition to front end
- Translations


js/components/resultsFormPC.js
```diff
@@ -0,0 +1,681 @@
...
+
+                <a 
+                href="#"
+                tabIndex="0"
+                data-toggle="popover"
+                data-placement="right"
+                data-trigger="focus"
+                data-content={ gettext('If data collection occurred within the fiscal year, enter the date where data was collected. If data collection occurred after the end of the fiscal year, enter the last day of the fiscal year (June 30).')}>
+                    &nbsp;<i className="far fa-question-circle"></i>
+                </a>
```
sanjuroj, line 270: With the exception of a couple of elements, these popover elements are fairly repetitious.  You should be able to build a separate component for these.  There may actually already be one available from the underlying program page.

js/components/resultsFormPC.js
```diff
@@ -0,0 +1,681 @@
...
+    }, [])
+
+    let handleDataEntry = (value, inputDisaggPk, customsort) => {
+        let update = {...disaggregationData};
+        update[inputDisaggPk].labels[customsort - 1] = {...disaggregationData[inputDisaggPk].labels[customsort - 1], value: value};
+        setDisaggregationData(update);
+    }
+
+    return (
+        <fieldset>
```
sanjuroj, line 472: Might be useful to have comments about which sections of the disaggregation input each one of these major divs represents.

js/components/resultsFormPC.js
```diff
@@ -0,0 +1,681 @@
...
+                        id="id_view_evidence_button" 
+                        className="btn btn-sm btn-secondary evidence-view__btn"
+                        disabled={!validEvidenceURL}
+                        onClick={() => window.open(evidenceFieldsInput.evidence_url, '_blank')}
+                    >{gettext('view')}</button>
+                    <button 
+                        type="button" 
+                        id="id_browse_google_drive" 
+                        className="btn btn-sm btn-link text-nowrap"
+                        onClick={() => alert("Google")}
```
sanjuroj, line 642: This button still needs to connect to google?

sanjuroj (changes requested):
Let's talk about the things that need to be changed.

PR #493, opened 2022-01-12
2647 result backend

sanjuroj:


sanjuroj:
André, this PR will:
- Add reporting period start and end dates to the result creation endpoint
- Make the result creation available to medium and high users, not just high users
- Fix the Python tests by rejiggering the way admin_type is delivered by the serializers

I'm hopeful that this will make the admin_type available to both program pages and IPTT pages.

andrethomas6 (approved):

PR #506, opened 2022-01-31
Updated translations from the translators.

andrethomas6:


sanjuroj (changes requested):
Looks like the temp translations are still there

sanjuroj (approved):
Sorry, looks fine, I was looking at deleted sections.

PR #513, opened 2022-02-02
Added participant count step by step guide to indicator form.

andrethomas6:


sanjuroj (approved):
My only suggestion would be to make the id a bit less generic.  Maybe `id_guidance_link--pc` instead of `id_link--pc`

andrethomas6:
@sanjuroj I made the id change and set english to the default link.

PR #521, opened 2022-02-11
Added backend validation for the PC results form

blakelong:
Added field validation to the PCResultSerializerWrite.

### DIFF
--- a/indicators/serializers_new/participant_count_serializers.py
+++ b/indicators/serializers_new/participant_count_serializers.py
@@ -1,6 +1,8 @@
 import copy
 import re
+from this import d
 from rest_framework import serializers, exceptions
+from django.utils.translation import ugettext_lazy as _
 from django.db.models import Count, BooleanField, Q
 from indicators.models import (
     PeriodicTarget, Result, OutcomeTheme, DisaggregationType, DisaggregationLabel, DisaggregatedValue
@@ -124,6 +126,16 @@ class PCResultSerializerWrite(serializers.ModelSerializer):
             'disaggregations'
         ]
 
+    def empty_evidence(self):
+        """
+        Utility method for checking if both evidence fields [evidence_url, record_name] are empty.
+
+        Returns
+            True if both fields are empty else False
+        """
+        return self.initial_data.get('record_name') == '' and self.initial_data.get('evidence_url') == '' \
+            or self.initial_data.get('record_name') is None and self.initial_data.get('evidence_url') is None
+
     def validate_date_collected(self, value):
         """
         Validates that date_collected is in the range of self.context.get('program').start_date and self.context.get('program').end_date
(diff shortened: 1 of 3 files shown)


indicators/serializers_new/participant_count_serializers.py
```diff
@@ -123,6 +124,93 @@ class Meta:
...
+
+        Returns
+            Value of date_collected
+        """
+        program = self.context.get('program')
+
+        if program.reporting_period_start <= value <= program.reporting_period_end:
+            return value
+
+        raise exceptions.ValidationError('This date should be within the fiscal year of the reporting period.')
```
sanjuroj, line 147: Any of these error messages that might be displayed by the frontend should be wrapped in translation tags and be preceded by translator comments.

sanjuroj (changes requested):
Generally, nice work, really clear code.  I made a couple of line-level comments.  The other thing that you probably want to add is validation for the disaggregation error messages.  Like, direct non-duplicated value needing to be the same as the SADD disaggregation, etc... André has a good handle on those since he just had to write them for the front-end.

indicators/serializers_new/participant_count_serializers.py
```diff
@@ -123,6 +124,93 @@ class Meta:
...
+        """
+        pattern = r"^(http(s)?|file):\/\/.+"
+        
+        if self.initial_data.get('record_name') is None or len(self.initial_data.get('record_name')) == 0:
+            raise exceptions.ValidationError('A record name must be included along with the link.')
+
+        if re.match(pattern, value) is None:
+            raise exceptions.ValidationError('Please enter a valid evidence link.')
+
+        return value
```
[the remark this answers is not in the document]
blakelong: I'll add a test case just to confirm, but I believe the validate_evidence_url method won't run if the evidence_url is not present. While testing locally I was able to add a result without supplying evidence.

indicators/serializers_new/participant_count_serializers.py
blakelong: Sounds good, I'll get those added.

indicators/serializers_new/participant_count_serializers.py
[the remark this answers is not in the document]
sanjuroj: Ah, gotcha, ok, if that's the case then not as much of an issue.  

However, I would definitely check out a regular result form to see if the  record name is required if there is an evidence url.  I don't believe it is.

indicators/serializers_new/participant_count_serializers.py
blakelong: I was looking at commits for translator comments. Would a translator comment be along the lines of `# Translators: An error message detailing that the selected date must be within the fiscal years reporting periods`

sanjuroj (approved):

PR #524, opened 2022-02-15
2661 tests for pc indicator create script

blakelong:
Added tests for the participant indicator create script. I had to omit the test case for test_without_create_disaggs_themes because the script was asking for input which failed on the test job when committing.

### DIFF
--- a/indicators/tests/test_management_create_participant_count_indicators.py
+++ b/indicators/tests/test_management_create_participant_count_indicators.py
@@ -1,0 +1,90 @@
+from django import test
+from django.core import management
+from datetime import date
+from indicators.models import Indicator, IndicatorType, OutcomeTheme, DisaggregationType, PeriodicTarget, DisaggregationLabel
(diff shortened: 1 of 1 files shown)


kbarkemeyer (commented):
Great clean tests!
The script also contains a --clean flag. Should we add a test for this as well?

blakelong:
> Great clean tests! The script also contains a --clean flag. Should we add a test for this as well?

I think we can skip the --clean flag. I tested it locally but was unable to add it to the test cases because it requires input from the user. I found out from the test case for --create_disaggs_themes=False that if the test case includes user input it will fail/freeze when GitHub processes the automatic test cases.

kbarkemeyer:
> > Great clean tests! The script also contains a --clean flag. Should we add a test for this as well?
> 
> I think we can skip the --clean flag. I tested it locally but was unable to add it to the test cases because it requires input from the user. I found out from the test case for --create_disaggs_themes=False that if the test case includes user input it will fail/freeze when GitHub processes the automatic test cases.

Thanks Blake! That makes sense! You all good to merge from my end!

PR #525, opened 2022-02-16
removed trailing whitespace to fix haiti bulk imports

blakelong:
Tracked down the bug to trailing whitespace around the Program.levels.leveltier. The whitespace was causing an if statement that checked if the level header name was a key in the level tier dictionary to evaluate as False.

Manually testing bulk imports on Haiti and other country programs passed.


kbarkemeyer (commented):
Awesome Blake! Good catch!

PR #528, opened 2022-02-23
Updated the total participant actual values helptext with extra note …

andrethomas6:
Updated the total participant actual values helptext with extra note on double counting.


kbarkemeyer (commented):
Looks good to me!

PR #530, opened 2022-02-24
subtracting 1 year from the fiscal year for program filtering

blakelong:
Including programs that ended last fiscal year into eligible_programs.


kbarkemeyer (commented):
Looks good!

PR #533, opened 2022-03-01
Added unique identifier to csv, synced data between csv and json views

blakelong:
Added unique identifier column to the CSV. Added utility function to keep the CSV and JSON data synced.


kbarkemeyer (commented):
Looks great Blake!

PR #535, opened 2022-03-07
2674 create pc indicator under additional circumstances

blakelong:
Added signal that triggers before a Program is saved in the database. The signal will create the participant count indicator if a Program changed from inactive to active and has the results framework.


kbarkemeyer (commented):
Looks great Blake!

kbarkemeyer (approved):

PR #536, opened 2022-03-07
2680 api errors for pc

andrethomas6:
- Added a server error component if the API fails
- Added a loading spinner component for when the pc results form opens up
- Prevent forever spinner after 1 min


blakelong:
Looks great!

kbarkemeyer (commented):
Looks good!

kbarkemeyer (approved):

PR #546, opened 2022-03-15
Updated named programs list for create qa programs command.

andrethomas6:
Updated named programs list for create qa programs command.


blakelong (approved):
Looks great!

PR #550, opened 2022-03-22
2737 update double counting heading text

andrethomas6:
- Changing the text of the Results form in the Total Participant Actual Values section from "With double counting" to "With double counting across programs"
- Updated Translations


blakelong (approved):
Looks great!

PR #554, opened 2022-03-31
For mercycorps/TolaActivity#2663, update for disabling fields for vie…

andrethomas6:
Update for disabling fields for view only state of PC Results Form.


blakelong (approved):
Looks great!

PR #557, opened 2022-04-05
Disabled save button on results framework while processing a request

blakelong:
Disabled the save button when creating a results framework while a request is processing.


kbarkemeyer (commented):
Looks great!

PR #564, opened 2022-04-11
Multi country programs now appear on separate rows. Added periodictarget to queryset filter

blakelong:
Multi-country programs appear in separate rows. Added PeriodicTarget filter to queryset.


kbarkemeyer (commented):
Looks good.

PR #566, opened 2022-04-14
updated function based program view to class based view

blakelong:
I updated the function-based program page view to a class-based view. The average response time went from 0.37 to 0.23 seconds, very minimal but more noticeable on slower internet connections.

### DIFF
--- a/indicators/views/views_program.py
+++ b/indicators/views/views_program.py
@@ -2,7 +2,8 @@
 """
     Program views: logframe, program page api, etc.
 """
-
+from django.views.generic import TemplateView
+from django.utils.decorators import method_decorator
 import math
 from operator import itemgetter
 import csv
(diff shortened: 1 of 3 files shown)


indicators/views/views_program.py
```diff
@@ -331,39 +332,52 @@ def get_parent_segment(ontology_segments, level):
...
+            'program': ProgramPageProgramSerializer.load_for_pk(self.program.pk).data,
+            'pinned_reports': list(PinnedReport.objects.filter(
+                program_id=self.program.pk, tola_user=self.request.user.tola_user
+                )) + [PinnedReport.default_report(self.program.pk)],
+            'delete_pinned_report_url': reverse('delete_pinned_report'),
+            'indicator_on_scope_margin': Indicator.ONSCOPE_MARGIN,
+            'levels': LevelSerializer(self.program.levels.all(), many=True).data,
+            'levelTiers': LevelTierSerializer(self.program.level_tiers.all(), many=True).data,
+            # 'levels': LevelSerializer(levels, many=True).data,
+            # 'levelTiers': LevelTierSerializer(tiers, many=True).data,
```
andrethomas6, line 361: Is this still needed?

andrethomas6 (approved):

indicators/views/views_program.py
blakelong: Whoops, I forgot to delete those. I'll re-push with the comment removed.

PR #568, opened 2022-04-20
For mercycorps/TolaActivity#2635 split outcome themes into their own columns

blakelong:
Split outcome themes from a single column to multiple.


kbarkemeyer (approved):
Looks good!

PR #572, opened 2022-04-29
%met now shows N/A on IPTT when last actual is 0

blakelong:
On the IPTT, when the last reported actual is 0, show N/A for the % met column. Added translation for N/A.


andrethomas6 (approved):
Did you also run "compilemessages" to update the .mo files?

blakelong:
> Did you also run "compilemessages" to update the .mo files?

Yeah, I reran it to be sure. The output was `/locale/fr/LC_MESSAGES/djangojs.po” is already compiled and up to date.` for both fr and es. I think the .mo files weren't updated because 'N/A' was already translated on that page.

PR #574, opened 2022-05-04
For mercycorps/TolaActivity#2778, Only show LOP targets and actuals r…

andrethomas6:
Only show LOP targets and actuals row for non Participant Count indicators on Program Page.


blakelong (approved):
Looks great!

PR #575, opened 2022-05-11
Save and link outcome button is now disabled while processing a request

blakelong:
The "Save and Link Outcome" button is now disabled while processing a request on the results framework page.


kbarkemeyer (approved):
Looks great!

PR #578, opened 2022-05-23
2663 pc creation script 2023

kbarkemeyer:
- Add script pc_indicators_add_periodic_target to add additional targets to existing pc indicators and change the customsort value for the periodic targets to fiscal year as int

- utils.py : 
Add method to add additional targets tp current pc inidcators
Add various method to create multiple targets when pc indicator is created
Add method to add or delete targets when program periods change

- signals.py:
Modify to compare current and instance program periods

- views_indicators.py:
Add methods to calculate reporting year and pt end and start date to be used when a result for an pc indicator is created.
Add viewonly variable to pt targets to ensure only current FY target can be modified

- indicators/models.py:
Add property is_fiscal_year to ensure only current FY targets can be modified

- participant_count_serializers.py:
Add view_only field to ensure results can only be added to/ be modified for current fiscal year

- views_results_framework.py
Make sure pc indicators are only created FY2022 and up

- Add appropriate variables in React
- Modify and add tests to accommodate changes


blakelong (approved):
Everything looks great! I liked how you used properties for the pt_start_date and pt_end_date. It was clever how the serializer used those.

PR #581, opened 2022-05-26
refactored gaitid and fundcode into a FK relationship with program table

blakelong:
* Refactored Program.gaitid to a new GaitID table. 
* Added management command to migrate existing gaitids to the new table.


kbarkemeyer (approved):
Great work. Nice, clean GaitID and FundCode classes! Thank you for creating the migration script and writing the tests!

PR #582, opened 2022-05-27
2785 update outcome themes and sectors

andrethomas6:
- Updated the sort/display order of the outcome themes.
- Added text wrapping for multiselect dropdown.


blakelong (approved):
Looks great!

PR #594, opened 2022-06-02
Validation class between IDAA and Tola programs

blakelong:
* Validation class to validate IDAA and Tola programs
* Sample IDAA JSON data
* Test cases


kbarkemeyer (approved):
Looks great! Very clever how you use class inheritance in program.py!

PR #596, opened 2022-06-07
2764 validation between tola and idaa programs

blakelong:
Updated the idaa test data structure and types to match idaa JSON response. Updated validation around GaitIDs to match the structure returned from idaa. Updated date format to match idaa date format.


kbarkemeyer (approved):
Thanks for doing this so quickly!

PR #601, opened 2022-06-16
Updated participant count actual and disaggregation value to allow fo…

andrethomas6:
Updated participant count actual and disaggregation value to 

- allow for values to be 0 (value of 0) or empty cells (value of null)
- removed requirement for indirect actuals values
- updated validation
- updated translations


blakelong (approved):
Looks great!

PR #602, opened 2022-06-16
Updated pc validation to allow 0 values

blakelong:
Updated pc validation to allow 0 values.

### DIFF
--- a/indicators/serializers_new/participant_count_serializers.py
+++ b/indicators/serializers_new/participant_count_serializers.py
@@ -297,9 +297,9 @@ class PCResultSerializerWrite(serializers.ModelSerializer):
             # Translators: An error message detailing that the sum of 'SADD without double counting' should be equal to the sum of 'Direct without double counting'
             raise exceptions.ValidationError(_("The sum of 'SADD without double counting' should be equal to the sum of 'Direct without double counting'."))
 
-        if actual_with_direct == 0 or actual_with_indirect == 0:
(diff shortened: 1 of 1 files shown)


kbarkemeyer (approved):
Great! Thank you!

indicators/serializers_new/participant_count_serializers.py
```diff
@@ -297,9 +297,9 @@ def disaggregations_are_valid(self, used_disaggregations):
             # Translators: An error message detailing that the sum of 'SADD without double counting' should be equal to the sum of 'Direct without double counting'
             raise exceptions.ValidationError(_("The sum of 'SADD without double counting' should be equal to the sum of 'Direct without double counting'."))
 
-        if actual_with_direct == 0 or actual_with_indirect == 0:
-            # Translators: An error message detailing that the fields Direct and Indirect total participants with double counting is required
-            raise exceptions.ValidationError(_("Direct/indirect total participants with double counting is required. Please complete these fields."))
+        if actual_with_direct is None or actual_with_indirect is None:
```
andrethomas6, line 300: Does this still raise an error if indirect is left empty or 0?

blakelong: It shouldn't raise an error. Empty values are handled before this method is called, and 0 values shouldn't raise the error. I tested with the frontend changes and could enter 0 or empty values.

PR #610, opened 2022-07-05
Connected program upload create and update

blakelong:
* Connected program upload create and update
* Bug fix around matching countries
* Error handling when the discrepancy report is missing a required value
* Moved country and gaitid API calls outside of ProgramUpload for performance
* Added new property ProgramUpload.program_updated to show if a program marked for an update had an update


kbarkemeyer (approved):
Thank you Blake! This looks great!

PR #611, opened 2022-07-06
Updated idaa country to reflect new idaa structure

blakelong:
Updated logic around IDAA countries to match the new structure returned from IDAA.


kbarkemeyer (approved):
Great! Thank you!

PR #613, opened 2022-07-12
Admin lite program history log shows changes made by IDAA

blakelong:
* Newly created programs display as created by IDAA
* Updated programs display as updated by IDAA
* Added management command to create a user with the username IDAA


kbarkemeyer (approved):
Thank you Blake, this looks great!

PR #617, opened 2022-07-14
Update README with installation instructions

axoplasm:
This is on my 2020-era model (Intel) MacBook Pro running MacOS 12.4 Monterey


blakelong (approved):
Looks great! Thank you for updating the readme! Since this is just the readme, there shouldn't be any merge conflicts. I think it would be ok to merge this into master but to merge master back down through the stack (staging, dev, transporter-3.0.0). Or if you want to re-point the PR into dev, that works too!

PR #621, opened 2022-07-18
2673 upgrade js libraries

axoplasm:
Narrowly addresses /mercycorps/tolaActivity#2673. This provides *updates* but not *upgrades, with the exception of `node-sass` → 7.0

Note also this does *not* remove the requirement to use node v.12, if you are like me using `nvm` remember to switch to v12 (`nvm use 12`) before updating packages or running `npm run watch`


andrethomas6 (approved):

blakelong (approved):
Looks great! I was able to run the build on my local.

kbarkemeyer (approved):
Works great. Just build it and ran it on my local!

PR #622, opened 2022-07-20
Replaced uglifyjs with terser plugin

blakelong:
Replaced uglifyjs with the terser plugin minify for production builds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,6 @@
         "select2": "^4.0.13",
         "style-loader": "^2.0.0",
         "topojson-client": "^3.1.0",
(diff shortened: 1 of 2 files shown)


kbarkemeyer (commented):
Should we also remove "uglifyjs-webpack-plugin": "^2.2.0" from the package.json file?

kbarkemeyer (approved):

blakelong:
> Should we also remove "uglifyjs-webpack-plugin": "^2.2.0" from the package.json file?

Oh yeah, I'll add a commit for that.

PR #626, opened 2022-07-21
2771 program period update backend

kbarkemeyer:
Created new api-view and serializers for program period React component
Added appropriate tests
Added new Program model property


blakelong (approved):
Looks great!

PR #627, opened 2022-07-25
Program admin lite gaitid changes

blakelong:
* Fixed GaitID error when creating a new program from the admin lite
* Updated admin lite JSON response to relate GaitIDs, fund codes, and donors
* Removed API call to Gait after a program is created


kbarkemeyer (approved):
Looks great!

PR #630, opened 2022-07-25
added donor_dept to gaitid response in program admin lite

blakelong:
Added donot_dept to gaitid response in program admin lite.


kbarkemeyer (approved):
Thanks Blake!

PR #631, opened 2022-07-27
Program upload outcome theme hotfix

blakelong:
Updated outcome theme structure to work with the structure sent from IDAA.


kbarkemeyer (approved):
Good catch! Thank you Blake!

PR #633, opened 2022-08-02
2673 upgrade js libraries (redux)

axoplasm:
Still dependent on npm v12

For mercycorps/TolaActivity#2673. See that issue for a play-by-play. Reinstates work in PRs #621, #622, #623, including replacing uglify with terser for production environments

Removes all critical vulnerabilities and will resolve all js-related Dependabot PRs

I have verified dev and prod builds on my local, but not on a server


blakelong (approved):
Looks great! I ran the following commands locally and did not encounter any errors.
`npm run install:dev` `npm run install:prod` `npm run build`

axoplasm:
This breaks on `dev2`. The villain seems to be that `render_bundle` doesn’t correct append `STATIC_URL`.

This may be because webpack configs are incorrectly specifying, or failing to specify, `publicPath` ??

See https://github.com/django-webpack/django-webpack-loader/issues/160

axoplasm:
This is obviated by #634

PR #638, opened 2022-08-08
2770 admin light profile tab updates

andrethomas6:
- Added program id, start date, end date, funding status, outcome themes, and gait / fund code / donors fields
- Kept some code in that will may be used in ticket 2820 that will make this form editable and able to update fields in the demo, devs, and localhost environments.

### DIFF
--- a/js/pages/tola_management_pages/program/components/edit_program_profile.js
+++ b/js/pages/tola_management_pages/program/components/edit_program_profile.js
@@ -25,9 +25,20 @@ export default class EditProgramProfile extends React.Component {
         const {program_data} = props
 
         this.state = {
+            formEditable: false,
             original_data: Object.assign({}, program_data),
             managed_data: Object.assign({}, program_data)
         }
+    }
+
+    componentDidMount() {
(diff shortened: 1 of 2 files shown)


kbarkemeyer (commented):
It seems that the outcome themes in Admin Lite are displayed incorrectly. The serializer is returning a list of [[ID, name]](https://mercycorps.slack.com/archives/D0344PJAF45/p1658864692591769), so both is displayed in the field instead of just the name. Would you like me to change the serializer to just return the name?

andrethomas6:
> It seems that the outcome themes in Admin Lite are displayed incorrectly. The serializer is returning a list of [[ID, name]](https://mercycorps.slack.com/archives/D0344PJAF45/p1658864692591769), so both is displayed in the field instead of just the name. Would you like me to change the serializer to just return the name?

Good catch. I just made a change that fixes it to only display the names. I'll still want the IDs to come in from the serializer so that when we do the update ticket 2820, I can send back the update with ids and names.

kbarkemeyer (approved):
Looks great! Thank you for changing this so quickly!

PR #639, opened 2022-08-09
For mercycorps/TolaActivity#2824 Fix sector issue in PCresults

kbarkemeyer:
Could you please pull my branch and double check that the server error is fixed before approving? (You might have to restore your local db in order to to this because I am branching off dev here.)


blakelong (approved):
Looks great! I was able to create and update results with sectors.

PR #653, opened 2022-08-22
Program upload create and fundcode fix

blakelong:
* Fix for resolved IDAA discrepancies showing on the discrepancy report.
* Fix for fund codes displaying as PK value on change log


kbarkemeyer (approved):
Good catch!

PR #654, opened 2022-08-23
Fixed issue where values for 'Peace, Governship and Partnership' not displaying

blakelong:
* Updated 'Governance and Partnership' to 'Peace, Governance and Partnership'


kbarkemeyer (approved):
Thank you for doing this so quickly!

PR #659, opened 2022-08-25
Fix for countries and sectors incorrectly showing as updated in status and history

blakelong:
Fix for countries and sectors incorrectly showing as updated in status and history.


kbarkemeyer (approved):
Great!

PR #661, opened 2022-08-26
Added list of created and updated programs to IDAA upload report

blakelong:
* Added list of programs created in TolaData to IDAA upload report
* Added list of programs updated in TolaData to IDAA upload report


kbarkemeyer (approved):
Great! Thank you Blake!

PR #663, opened 2022-08-30
For mercycorps/TolaActivity#2770, display none for null idaa program …

andrethomas6:
Display `None` for null idaa program id field.


blakelong (approved):
Looks great!

PR #665, opened 2022-08-31
added idaa_sectors to program profile page. Updated sector filters to idaa_sectors

blakelong:
* Added idaa_sectors to the admin lite program profile page
* Updated sector filters on the admin lite program page to use the idaa_sectors


kbarkemeyer (approved):
Looks great!

PR #669, opened 2022-09-06
Fix PCIndicator outcome_theme and sector validation bugs

kbarkemeyer:
- Rewrite PC Indicator Result update method, so outcome themes can be added and removed.
- Change PC Indicator Result disaggregation validations, so a 'no input' in Actuals will not cause a server error when sector values are recorded as 0.

For https://github.com/mercycorps/TolaActivity/issues/2838 and https://github.com/mercycorps/TolaActivity/issues/2836


blakelong (approved):
Looks great!

PR #681, opened 2022-09-19
2821 update create qa programs

kbarkemeyer:
For For https://github.com/mercycorps/TolaActivity/issues/2821, update create-qa-programs management script


blakelong (approved):
Looks great!

PR #682, opened 2022-09-20
2767 existing program updates

blakelong:
Changes for https://github.com/mercycorps/TolaActivity/issues/2764#issuecomment-1251509113.

* One IDAA program to many TolaData programs will not update the TolaData program name
* Fund codes in TolaData are deleted if deleted in IDAA
* Valid tracking dates method checks the range from the 1st and last day of the month instead of the program start and end date.
* Prevent multiple_program discrepancies from having other discrepancies
* Fixed issue where not all discrepancies were captured for invalid IDAA programs
* Updated test on discrepancy report
* Fixed test cases
* Fixed KeyError when logging an invalid IDAA program


kbarkemeyer (approved):
Great!

PR #686, opened 2022-09-28
fixed issue where sectors and outcome themes are deleted from the DB

blakelong:
Fixed issue where sectors and outcome themes are deleted from the DB.


kbarkemeyer (approved):
Good catch!

PR #689, opened 2022-09-28
In Django admin: all create/edit date fields are read-only & appear in list views

axoplasm:
for mercycorps/TolaActivity#2879.

This one has several changes, would appreciate verification of all screens from at least one other local development box. See original ticket for a list of models


kbarkemeyer (approved):

blakelong (approved):
Looks great! I confirmed the create and edit dates were read-only on each screen. I noticed that some models define the create and edit date but don't have the fields populated automatically or manually (IndicatorType, ExternalServiceRecord, ExternalServices, DataCollectionFrequency).

PR #691, opened 2022-09-29
Management command to alias all country admin emails

blakelong:
For mercycorps/TolaActivity#2881

Added a new management command `alias_country_admin_emails` to alias every country admin email with `tola+`

After commit https://github.com/mercycorps/toladata/pull/691/commits/386a4029544efcf23ddce7db6bb46fe318e8e53e

* Moved command to /tola/management/commands directory
* Added regex match to check if an email has already been aliased
* Alias all user emails instead of country admin emails
* Print output to `STDOUT`


kbarkemeyer (approved):
Looks great!

workflow/management/commands/alias_country_admin_emails.py
```diff
@@ -0,0 +1,68 @@
...
+        )
+        parser.add_argument(
+            '--supress_output', action='store_true', help='Hide text output'
+        )
+
+    def get_country_admins(self):
+        return CountryAccess.objects.filter(role='basic_admin').exclude(tolauser__user__is_superuser=True)
+
+    def slugify_non_mc_email(self, email):
+        chars = ['@', '.']
```
axoplasm, line 24: also `+`

workflow/management/commands/alias_country_admin_emails.py
```diff
@@ -0,0 +1,68 @@
...
+        return email
+
+    def handle(self, *args, **options):
+        """
+        Loops through all country admins and prefixes each email with tola+
+        """
+        emails_aliased = 0
+
+        if not options['supress_output'] and not options['execute']:
+            print('Missing execute flag. This will be a dry run.')
```
axoplasm, line 44: Without looking at the source I don’t know what happens during a “dry run” (nothing, apparently?). One thing we could is pipe the output (or first `n` lines) to STDOUT

axoplasm (changes requested):
When I ran this command on my local, it only aliased 120 users. I think it hit a character other than `@` or `.` and exited. __Needs more testing__

Some suggestions:

- Default verbose output to `STDOUT`
- Should we use an external slugify library e.g. `[python-slugify](https://pypi.org/project/python-slugify/)`?

axoplasm:
Minor nit: because this command affects multiple apps (`auth` and `workflow`) it should probably live in the project app directory ie. `/tola/management/commands`

workflow/management/commands/alias_country_admin_emails.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Prevent the script from running on production
+        if socket.gethostname() != self.production_host_name:
+            country_admins = self.get_country_admins()
```
axoplasm, line 48: I don’t think we need to get country admins when what we really want to change are *all* the user emails. `get_country_admins()` will produce a lot of duplicates, e.g. when a user is an admin for multiple countries.

workflow/management/commands/alias_country_admin_emails.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        Loops through all country admins and prefixes each email with tola+
+        """
+        emails_aliased = 0
+
+        if not options['supress_output'] and not options['execute']:
+            print('Missing execute flag. This will be a dry run.')
+
+        # Prevent the script from running on production
+        if socket.gethostname() != self.production_host_name:
```
axoplasm, line 47: this becomes a pretty deeply nested loop. Couple suggestions:

- Reverse the `if` on this line: exit the loop entirely if we are on production
- No need for the `for` statement on line 49 (see note below)
- move the check on line 52 into a function (see note below)

workflow/management/commands/alias_country_admin_emails.py
```diff
@@ -0,0 +1,68 @@
...
+            country_admins = self.get_country_admins()
+            for country_admin in country_admins:
+                user = country_admin.tolauser.user
+                # Check that the email has not already been aliased
+                if user.email[0:len(self.alias_email)] != self.alias_email:
```
axoplasm, line 52: I think I’d do 2 things here:

1. move this check into a separate function to clean up this loop
2. check against the `tola+[...]@mercycorps.org` pattern, so we can safely run this command multiple times (otherwise we might get into a situation where we have a bunch of addresses like `tola_tola+tola+fake_address@mercycorps.org`)

axoplasm (changes requested):

axoplasm:
@blakelong can you see the micromentor repo esp. [this command](https://github.com/mercycorps/micromentor/blob/main/core/management/commands/disable_external_user_accounts.py)? The gist of how we did it there:

```python
from django.db.models import Q, F, Value
from django.db.models.functions import Concat, Replace
from django.core.management.base import BaseCommand
from core import models

# ...

class Command(BaseCommand):
    def handle(self, *args, **options):

        user_ids_to_exclude = list(models.User.objects.filter(
            Q(is_mm_staff=True) |
            Q(is_admin=True) |
            Q(email__icontains='%@micromentor.org') |
            Q(email__icontains='%@mercycorps.org')
        ).filter(is_active=True).values_list('id', flat=True))

        # ...

        print('Setting user emails to @example.com')
        # update core_user set email = replace(email, '@', '+') || '@example.com' where email not ilike '%@micromentor.org' and email not ilike '%@example.com';
        models.User.objects.exclude(id__in=user_ids_to_exclude).exclude(email__icontains='%@example.com').update(
            email=Concat(
                Replace(F('email'), Value('@'), Value('+')),
                Value('@example.com')
            )
        )

```

workflow/management/commands/alias_country_admin_emails.py
kbarkemeyer: @axoplasm I am confused. I thought the ticket specifies to only change the country admin emails. Should we change the ticket and change all user emails?

PR #695, opened 2022-10-03
Log removing old sectors. Translations for IDAA changes history log

blakelong:
* When updating the upload_IDAA_programs management command runs, remove and log old sectors from the program
* Fix issue where empty fields on the History log shows up as commas
* Translations for the History log changes

For mercycorps/TolaActivity#2772


kbarkemeyer (approved):
Looks good!

PR #697, opened 2022-10-03
2821 update create qa programs

kbarkemeyer:
- Fix mistake in demo one user profile
- Comment out creation of pre-Satsuma program (no need since there is now new program without rf)
- Add hard coded dates to past program (so pc indicator does not get added)


blakelong (approved):
Looks great! I ran the command on my local and confirmed:
* demo one has the organization Test
* Ghost of Programs Past does not have a PC indicator
* The pre-satsuma program was not created

PR #698, opened 2022-10-04
remove fundcodes from the tola program

blakelong:
remove fundcodes from the tola program

### DIFF
--- a/workflow/program.py
+++ b/workflow/program.py
@@ -670,20 +670,25 @@ class ProgramUpload(ProgramValidation):
             gaitid_obj.save()
 
             if 'FundCode' in gaitid_details:
-                idaa_fund_codes = gaitid_details['FundCode'].split(',')
+                idaa_fund_codes = []
+
+                for idaa_fund_code in gaitid_details['FundCode'].split(','):
+                    try:
+                        idaa_fund_codes.append(int(idaa_fund_code))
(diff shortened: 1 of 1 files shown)


kbarkemeyer (commented):
I am not quite sure I understand the logic of turning x into string in BOTH lists. IDAA fundcodes come through as strings, so it makes sense to turn them into strings in the second list. But tola fundcodes are integers so shouldn't we turn the idaa fundcodes into integers in the first list?
Also, do you think it might be safer to turn the IDAA strings into integers before we create?

blakelong:
@kbarkemeyer 
> I am not quite sure I understand the logic of turning x into string in BOTH lists. IDAA fundcodes come through as strings, so it makes sense to turn them into strings in the second list. But tola fundcodes are integers so shouldn't we turn the idaa fundcodes into integers in the first list?

Ah, that's right, I missed that I can switch it.

>Also, do you think it might be safer to turn the IDAA strings into integers before we create?

Originally I skipped that because IDAA sends some fundcodes that cant be converted to an int, and we were catching those when creating the fundcode. I agree it would be safer to convert to an int. I'll update it to convert the IDAA fundcodes to an int before creating the lists and move the try-except to that level.

kbarkemeyer (approved):
Looks great!

PR #704, opened 2022-10-06
Change max_length for IDAAOutcomeTheme.name to 255

kbarkemeyer:
[Django Docs](https://docs.djangoproject.com/en/3.2/ref/databases/#mysql-notes)

Screen Shot 2022-10-06 at 1 21 40 PM


blakelong (approved):
Good catch!

PR #705, opened 2022-10-07
translations for the country created history log

blakelong:
Translations for the country created history log

For mercycorps/TolaActivity#2876


kbarkemeyer (approved):
Looks good!

PR #706, opened 2022-10-07
Editable GaitID form

axoplasm:
For mercycorps/TolaActivity#2886.

Either Karen or Blake can review. Note there is a migration in this PR!


axoplasm:
This also fixes a bug where `GaitID.fund_code()` returned an empty list `[]` as it was references objects that were refactored away. It now returns a list of Fund Codes (not `FundCode` objects), which I believe was the original behavior

workflow/admin.py
```diff
@@ -87,6 +87,13 @@ def formfield_for_foreignkey(self, db_field, request=None, **kwargs):
 
 class GaitIDInlineAdmin(admin.StackedInline):
     model = GaitID
+    fields = ['gaitid', 'donor', 'donor_dept']
+    readonly_fields = ['fund_codes',]
```
blakelong, line 91: On the Program page showing the `GaitIDInlineAdmin`, I do not see the fund_codes. I think the `fields` list needs `fund_codes` added to it as well.

workflow/admin.py
```diff
@@ -87,6 +87,13 @@ def formfield_for_foreignkey(self, db_field, request=None, **kwargs):
...
+    extra = 0
```
blakelong, line 92: On the Program page showing the `GaitIDInlineAdmin`, I don't see the fund_codes listed. I think the `fields` list needs `fund_codes` added.

blakelong (approved):
Looks great! I like how you added the FundCodes to the GaitID page.

workflow/admin.py
axoplasm: Good catch Blake! Unfortunately I don’t think we can put an inline inside an inline 😭

PR #708, opened 2022-10-12
2884 allow empty fields programadmin serializer

kbarkemeyer:
- Added attributes to NestedGaitIDSerializer fields to allow for null or None values in donor and donor_dept fields
- Removed unused fields from ProgramAdminSerializer
- Added attributes to fields in ProgramAdminSerializer to allow for null or None values for fields that are not required.
- Removed and changed validations.
- Update the create and update methods accordingly.
@blakelong I realized that the donor_dept had not been saved by the serializer previously and added that. I assume this was a bug and not designed this way for some reason?
- Added unit tests.

### DIFF
--- a/tola_management/programadmin.py
+++ b/tola_management/programadmin.py
@@ -257,8 +257,8 @@ class NestedFundCodeSerializer(ModelSerializer):
 
 class NestedGaitIDSerializer(ModelSerializer):
     gaitid = IntegerField(max_value=99999)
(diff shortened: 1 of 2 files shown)


blakelong (approved):
Looks great! 

Good catch on the `donor_dept`. I initially omitted it because we weren't displaying a field. But with the updates to the Acceptance Criteria for mercycorps/TolaActivity#2820, we are displaying an editable field.

PR #713, opened 2022-10-17
2820 editable program admin light

andrethomas6:
2820 editable program admin light


kbarkemeyer (commented):
I am trying out the changes on my local and everything looks great. One little issue with the validations: If I go into an old program profile tab with missing (required) values, just change one of the required fields, and then try to save, it doesn't save but not warning message appears either. For example I go into an old program with missing external ID and just change the Gait ID, I cannot save, but I also do not get a message about the other required fields.) Could you look into this?

axoplasm (changes requested):
I hit a serious bug with the Countries dropdown when testing this on my local.

axoplasm:
Copied from 2820, I was treating this like Acceptance Review when really it isn't! 

----

### Testing notes pt. 1

- *Serving on localhost:8000*
- *Using Safari 16.0, Firefox 104.0.2, and Firefox 106.0 (all on Mac OS 12.6)*
- *Tested in English only*

#### Dev notes:

- I needed to run `npm install` before testing
-  …this also bumped jQuery (`"jquery": ">= 2.2.4"` & required bootstrap (`"bootstrap": "~4.5.x",`) in my `package-lock.json ` @andrethomas6 : __should we commit this lock?__

#### Notes re: acceptance criteria

- When I supply a start date but not an end date, both fields are marked as incomplete and there is no error message:
image
- If I scroll through the Countries list, the list is truncated in the Ms or Ns, depending on how fast I scroll: 
image … however Countries beginning with later letters are still available when I type those letters 
image
- When I return to edit the same Program, if I have selected a country that is below the truncation, I cannot unselect it:
image

#### UX notes:
- Error handling occurs on Submit…do we do this realtime (asynchrnously) elsewhere? What is our *preferred* method for error checking?
- “Add another row” lacks context (what row am I adding?) Should it read “Add another GAIT ID”?

kbarkemeyer (commented):
I ran into the same issues Paul did on my local in Chrome, except I am able to unselect previously selected country.

axoplasm:
The Countries bug is not blocking for this PR, but I'd like to address the other two issues I found:

1. I needed to run `npm install` before testing…this also bumped jQuery (`"jquery": ">= 2.2.4"` & required bootstrap (`"bootstrap": "~4.5.x",`) in my `package-lock.json ` @andrethomas6 : __should we commit this lock?__
2. When I supply a start date but not an end date, both fields are marked as incomplete and there is no error message:
image

andrethomas6:
> I am trying out the changes on my local and everything looks great. One little issue with the validations: If I go into an old program profile tab with missing (required) values, just change one of the required fields, and then try to save, it doesn't save but not warning message appears either. For example I go into an old program with missing external ID and just change the Gait ID, I cannot save, but I also do not get a message about the other required fields.) Could you look into this?

@kbarkemeyer This should be fixed now.

andrethomas6:
> The Countries bug is not blocking for this PR, but I'd like to address the other two issues I found:
> 
> 1. I needed to run `npm install` before testing…this also bumped jQuery (`"jquery": ">= 2.2.4"` & required bootstrap (`"bootstrap": "~4.5.x",`) in my `package-lock.json ` @andrethomas6 : **should we commit this lock?**
> 2. When I supply a start date but not an end date, both fields are marked as incomplete and there is no error message:
>    image

@axoplasm I updated and fixed this problem with the dates. It should be working as expected now.

kbarkemeyer (commented):
The validations in the program profile page look great, but the reason that the errors on the fields prior to Gait ID are not showing seems to be that the errors are only displayed on input. Here is the example for the external program id:

Screen Shot 2022-10-18 at 1 25 44 PM

If I never click this field, the error will never be displayed.

kbarkemeyer:
> > I am trying out the changes on my local and everything looks great. One little issue with the validations: If I go into an old program profile tab with missing (required) values, just change one of the required fields, and then try to save, it doesn't save but not warning message appears either. For example I go into an old program with missing external ID and just change the Gait ID, I cannot save, but I also do not get a message about the other required fields.) Could you look into this?
> 
> @kbarkemeyer This should be fixed now.

@andrethomas6 Sorry I didn't your latest commit. Disregard my last comment. It works perfectly now!

kbarkemeyer (approved):

blakelong (approved):
Looks great! I tested on my local, and everything is working!

axoplasm:
Thanks for fixing the datefield error thing I noted above.

I went ahead and committed the updated `package-lock.json` but it’s strictly housekeeping — these are updated deps for bootstrap and jquery which are both overridden elsewhere.

As soon as we resolve conflicts in datestamps for translation files then this is OK to merge.

axoplasm (approved):

PR #717, opened 2022-10-20
Removed React Datepicker and added old datepicker for start and end d…

andrethomas6:
Removed React Datepicker and added old datepicker for start and end dates.
Fixed outcome themes hover text.


blakelong (approved):
Looks great! I was able to run the build without errors.

PR #722, opened 2022-10-25
Update to fix translating funding status.

andrethomas6:
For mercycorps/TolaActivity#2770, update to fix translating funding status.


kbarkemeyer (approved):
Great, thank you!

PR #727, opened 2022-11-02
Indicator json hotfix

kbarkemeyer:


blakelong (approved):
Works great! I tested locally with each indicator that sent an error.

